PR #191, opened 2018-11-01
"authz_domain" is an optional field in the JSON configuration file.

falfaro:
PR #177 introduced the "authz_domain" field in the JSON configuration file which means that BKPR deployments prior to PR #177 won't have it listed among the attributes. Hence this PR provides a default value of "*" when it is not defined.


anguslees:
.. you should re-run `kubeprod install ...` - no compatibility guarantees before 1.0 ;)

Issue #193, opened 2018-11-01
Tons of fluentd logs brings down the ES cluster by disk exhaustion 


falfaro:
On a completely idle GKE cluster, and after a week, some ElasticSearch pods have entered into a crash loop because the underlying PV is full.

The ElasticSearch manifest uses 100GB by default for each PV. At the moment only one of the three ElasticSearch pods is crashing, but the other two pods will very soon start crashing as they PVs are already at 100% according to "df -k":

```
$ kubectl --namespace=kubeprod exec -it elasticsearch-logging-1 bash
Defaulting container name to elasticsearch-logging.
Use 'kubectl describe pod/elasticsearch-logging-1 -n kubeprod' to see all of the containers in this pod.
I have no name!@elasticsearch-logging-1:/$ df -k
Filesystem     1K-blocks     Used Available Use% Mounted on
overlay         98868448  6814184  92037880   7% /
tmpfs            3829072        0   3829072   0% /dev
tmpfs            3829072        0   3829072   0% /sys/fs/cgroup
/dev/sda1       98868448  6814184  92037880   7% /etc/hosts
shm                65536        0     65536   0% /dev/shm
/dev/sdb       102687672 97428408         0 100% /bitnami/elasticsearch/data
tmpfs            3829072       12   3829060   1% /run/secrets/kubernetes.io/serviceaccount
tmpfs            3829072        0   3829072   0% /sys/firmware
```

arapulido:
It seems that this needs more investigation. The PV shouldn't be filled up that quickly

jbianquetti-nami:
Fluentd data fills data volumes of ES. 

That's caused by  fluent container settings which set rules to monitor all containers STDIN/STOUT/log files, including himself, and therefore causing a write-read-write-read loop on log files.

related PR: https://github.com/bitnami/kube-prod-runtime/pull/205

falfaro:
It is not enough with changing `fluentd` to only log `ERROR` and `FATAL`. We will probably need to implement some pre-processing to ignore log entries from the `fluentd` containers in the first place.

jjo:
Went down comparing our internal fluentd-es configs with BKPR ones, finding
nothing relevant ...  but then realized we're using an upstream image instead
(`gcr.io/google-containers/fluentd-elasticsearch:v2.0.1`) which nicely comes with the
following **embedded** `/etc/fluent/fluent.conf`:
```
# This is the root config file, which only includes components of the actual configuration
[...]

PR #205, opened 2018-11-01
Reduce `@log_level` to `error` for fluentd.

falfaro:
The goal is to avoid fluentd logging to `stdout` to have those logs
collected by itself again and streamed to Elasticsearch, eventually
exhausting Elasticsearch storage capacity.


manifests/components/fluentd-es-config/output.conf
```diff
@@ -6,23 +6,21 @@
...
+  # Set the chunk limits.
+  buffer_chunk_limit 2M
+  buffer_queue_limit 8
+  # Bitnami: 5s -> 30s
+  flush_interval 30s
+  # Never wait longer than 5 minutes between retries.
+  max_retry_wait 30
+  # Bitnami: retry 5 times most instead of forever (our elasticsearch deploy is capped to 2G mem)
+  disable_retry_limit false
+  retry_limit 5
```
anguslees: Does this mean we try 5 times, and then lose the logs?
(that's ok, I just want to understand the implications and overall timeout behaviour)

manifests/components/fluentd-es-config/output.conf
```diff
@@ -6,23 +6,21 @@
...
-    queue_limit_length 8
-    overflow_action block
-  </buffer>
+  # Set the chunk limits.
+  buffer_chunk_limit 2M
+  buffer_queue_limit 8
+  # Bitnami: 5s -> 30s
+  flush_interval 30s
+  # Never wait longer than 5 minutes between retries.
+  max_retry_wait 30
```
anguslees: I don't think the comment matches the parameter value.  `300` (5min) vs `30`

jjo:
See my comment at https://github.com/bitnami/kube-prod-runtime/issues/193#issuecomment-436266408,
adding that configfile with `fluent.**` filtering-out should solve the issue, and let's please
have this PR only addressing this one.

arapulido:
Thanks for the help @jjo !

PR #211, opened 2018-11-01
Get kubeprod latest release programatically in quickstart guides

dbarranco:
This will default to the latest *release* available of kubeprod. By the moment it will be set to:

```
~/projects/kube-prod-runtime$ echo $BKPR_VERSION
v0.3.1
```

Let me know what you think :)

### DIFF
--- a/docs/quickstart-aks.md
+++ b/docs/quickstart-aks.md
@@ -85,10 +85,8 @@ BKPR releases are available for 64-bit versions of Linux, macOS and Windows plat
 For convenience lets define a environment variable with the BKPR version:
 
 ```bash
-export BKPR_VERSION=vX.Y.Z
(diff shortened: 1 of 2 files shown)


docs/quickstart-aks.md
```diff
@@ -85,11 +85,9 @@ BKPR releases are available for 64-bit versions of Linux, macOS and Windows plat
 For convenience lets define a environment variable with the BKPR version:
 
 ```bash
-export BKPR_VERSION=vX.Y.Z
+export BKPR_VERSION=$(curl --silent "https://api.github.com/repos/bitnami/kube-prod-runtime/releases/latest" | sed -n 's/.*"tag_name": *"\([^"]*\)".*/\1/p')
```
sameersbn: since the guide lists `jq` as a requirement, we could use `jq -r '.tag_name'` instead of the `sed` command.

dbarranco:
done, thanks for the feedback 👍

PR #214, opened 2018-11-05
GKE: Enable integration tests

sameersbn:
- [x] Configure CI setup for testing on GKE clusters
- [x] Configure DNS suffix for GKE tests
- [x] Clean up disk leakage
- [x] Debug monitoring test failures in GKE

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -242,7 +242,7 @@ spec:
                             // $AZURE_SUBSCRIPTION_ID, $AZURE_TENANT_ID.
                             withCredentials([azureServicePrincipal('jenkins-bkpr-owner-sp')]) {
                                 def resourceGroup = 'jenkins-bkpr-rg'
-                                def clusterName = ("${env.BRANCH_NAME}".take(8) + "-${env.BUILD_NUMBER}-" + UUID.randomUUID().toString().take(5) + "-${platform}").replaceAll(/[^a-zA-Z0-9-]/, '-').toLowerCase()
+                                def clusterName = ("${env.BRANCH_NAME}".take(8) + "-${env.BUILD_NUMBER}-" + UUID.randomUUID().toString().take(5) + "-${platform}").replaceAll(/[^a-zA-Z0-9-]/, '-').replaceAll(/--/, '-').toLowerCase()
                                 def dnsPrefix = "${clusterName}"
                                 def parentZone = 'tests.bkpr.run'
                                 def dnsZone = "${dnsPrefix}.${parentZone}"
@@ -374,7 +374,7 @@ az account set -s $AZURE_SUBSCRIPTION_ID
                                 dir('src/github.com/bitnami/kube-prod-runtime') {
                                     def project = 'bkprtesting'
                                     def zone = 'us-east1-d'
-                                    def clusterName = ("${env.BRANCH_NAME}".take(8) + "-${env.BUILD_NUMBER}-" + UUID.randomUUID().toString().take(5) + "-${platform}").replaceAll(/[^a-zA-Z0-9-]/, '-').toLowerCase()
+                                    def clusterName = ("${env.BRANCH_NAME}".take(8) + "-${env.BUILD_NUMBER}-" + UUID.randomUUID().toString().take(5) + "-${platform}").replaceAll(/[^a-zA-Z0-9-]/, '-').replaceAll(/--/, '-').toLowerCase()
                                     def dnsPrefix = "${clusterName}"
                                     def parentZone = 'tests.bkpr.run'
                                     def adminEmail = "${clusterName}@${parentZone}"
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -282,6 +282,7 @@ az aks create                      \
 
                                             sh "az aks get-credentials --name ${clusterName} --resource-group ${resourceGroup} --admin --file \$KUBECONFIG"
 
+
                                             // create dns zone
                                             sh "az network dns zone create --name ${dnsZone} --resource-group ${resourceGroup} --tags 'platform=${platform}' 'branch=${BRANCH_NAME}' 'build=${BUILD_URL}'"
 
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -335,6 +356,99 @@ az account set -s $AZURE_SUBSCRIPTION_ID
...
+                                            ])
+                                        }
+                                    }
+                                    finally {
+                                        container('gcloud') {
+                                            def disksFilter = "${clusterName}".take(18).replaceAll(/-$/, '')
+                                            sh "gcloud auth activate-service-account --key-file ${GOOGLE_APPLICATION_CREDENTIALS}"
+                                            sh "gcloud container clusters delete ${clusterName} --zone ${zone} --project ${project} --quiet || :"
+                                            sh "gcloud compute disks delete \$(gcloud compute disks list --project ${project} --filter name:${disksFilter} --format='value(name)') --project ${project} --zone ${zone} --quiet || :"
+                                            sh "gcloud dns managed-zones delete \$(gcloud dns managed-zones list --filter dnsName:${dnsZone} --format='value(name)' --project ${project}) --project ${project} || :"
```
anguslees: Aside: If it becomes relevant at some point, we can easily delete all the entries in a managed zone using:
```
gcloud dns record-sets import /dev/null --zone=$zone --delete-all-existing
```

Jenkinsfile
```diff
@@ -335,6 +356,99 @@ az account set -s $AZURE_SUBSCRIPTION_ID
...
+                                    try {
+                                        runIntegrationTest(platform, "gke --project=${project} --dns-zone=${dnsZone} --email=${adminEmail} --authz-domain=\"*\"", "") { // --dns-suffix ${dnsZone}
+                                            container('gcloud') {
+                                                sh "gcloud auth activate-service-account --key-file ${GOOGLE_APPLICATION_CREDENTIALS}"
+                                                sh """
+gcloud container clusters create ${clusterName} \
+ --cluster-version ${kversion} \
+ --project ${project} \
+ --create-subnetwork '' \
+ --enable-ip-alias \
```
anguslees: I think we should remove `--create-subnetwork "" --enable-ip-alias` ?
I can't remember why I put them there (anticipating one of the many future-compatibility-warnings that `gcloud c c create` reports?), and they aren't mentioned in the gke quickstart.  Better to test whatever we recommend in the docs.

Jenkinsfile
```diff
@@ -224,15 +242,15 @@ spec:
                             // $AZURE_SUBSCRIPTION_ID, $AZURE_TENANT_ID.
                             withCredentials([azureServicePrincipal('jenkins-bkpr-owner-sp')]) {
                                 def resourceGroup = 'jenkins-bkpr-rg'
-                                def clusterName = ("${env.BRANCH_NAME}".take(8) + "-${env.BUILD_NUMBER}-${platform}-" + UUID.randomUUID().toString().take(5)).replaceAll(/[^a-zA-Z0-9-]/, '-').toLowerCase()
+                                def clusterName = ("${env.BRANCH_NAME}".take(8) + "-${env.BUILD_NUMBER}-" + UUID.randomUUID().toString().take(5) + "-${platform}").replaceAll(/[^a-zA-Z0-9-]/, '-').toLowerCase()
```
anguslees: We need more uniqueness than platform+BUILD_NUMBER?  This expression is getting a bit silly - I think if we're going to include a random component, then lets just drop all the non-random bits and rely on tags/labels to report meta-info.

tests/monitoring_test.go
```diff
@@ -87,6 +87,11 @@ var _ = Describe("Monitoring", func() {
 	var ns string
 
 	BeforeEach(func() {
+		if *dnsSuffix == "" {
```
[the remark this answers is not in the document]
sameersbn: That what I thought, but for some reason the service endpoint is not reachable via the proxy

Jenkinsfile
sameersbn: The truncation removes the uniqueness. This change essentially is to make sure the uniqueness does not get removed as a result of truncations

Jenkinsfile
sameersbn: yep I will remove this. It's not needed

tests/monitoring_test.go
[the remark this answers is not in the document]
sameersbn: for reference here's one of the instances where the monitoring tests failed without a valid dns setup https://jenkins-bkpr.k.dev.bitnami.net/blue/organizations/jenkins/kube-prod-runtime/detail/private-sbox/35/pipeline/83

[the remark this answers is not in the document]
sameersbn: well.. it seems the error is not related to the DNS setup. I got it wrong. Need to find out what is causing the monitoring tests to fail

Jenkinsfile
sameersbn: thanks for this. Using this now and reverted back to deleting the cluster asynchronously.

Jenkinsfile
sameersbn: cleaned up

tests/monitoring_test.go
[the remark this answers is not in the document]
sameersbn: removed this condition. debugging the reason for monitoring test failures on GKE. Oddly I don't notice any issues while running the tests locally

Jenkinsfile
```diff
@@ -26,15 +29,70 @@ def withGo(Closure body) {
...
+        }
+    }
+}
+
+def insertGlueRecords(String name, String nameServersJson, String ttl, String zone, String resourceGroup) {
+    withCredentials([azureServicePrincipal('jenkins-bkpr-owner-sp')]) {
+        container('az') {
+            sh "az login --service-principal -u \$AZURE_CLIENT_ID -p \$AZURE_CLIENT_SECRET -t \$AZURE_TENANT_ID"
+            sh "az account set -s \$AZURE_SUBSCRIPTION_ID"
+            for (ns in readJSON(text: nameServersJson)) {
```
anguslees: Minor: Probably makes sense to make the `nameServers` argument to this function be an array/list of strings and call `readJSON` in the caller(s).  So far 2 out of 2 give it to us as a json-serialised array, so I'm also ok with leaving that as the "native form" for now...

PR #220, opened 2018-11-06
Do not collect fluentd's own logs to avoid infinite loops.

falfaro:
Also bump `fluentd` from 1.2.2 to 1.2.6.


manifests/components/fluentd-es.jsonnet
```diff
@@ -21,7 +21,7 @@ local kube = import "../lib/kube.libsonnet";
 local kubecfg = import "kubecfg.libsonnet";
 local utils = import "../lib/utils.libsonnet";
 
-local FLUENTD_ES_IMAGE = "bitnami/fluentd:1.2.2-r22";
+local FLUENTD_ES_IMAGE = "bitnami/fluentd:1.2.6-r16";
```
sameersbn: we'll want to backport this change. It would help the case if the version bump can be avoided

falfaro:
Folks, the approach by @jjo doesn't work for some reason, and I also don't like the idea of discarding `fluentd` logs to /dev/null. Instead, I'm instructing `fluentd` to send error or fatal logs to `/tmp/fluentd.log` instead of standard output and use log fluentd built-in log rotation capabilities.

falfaro:
I found that the Docker image built-in `fluentd.conf` makes some assumptions that produce the infinite logging loop. Instead, we are using a custom `fluentd.conf` file that just includes all the configuration files from the `config.d` directory.

Could you PTAL?

manifests/components/fluentd-es.jsonnet
```diff
@@ -80,7 +86,7 @@ local FLUENTD_ES_LOG_BUFFERS_PATH = "/var/log/fluentd-buffers";
                 runAsUser: 0,  // required to be able to read system-wide logs.
               },
               env_+: {
-                FLUENTD_OPT: "-q",
+                FLUENTD_OPT: "-o /tmp/fluentd.log --log-rotate-age 5 --log-rotate-size 104857600 --no-supervisor",
```
sameersbn: i would suggest dropping the logs in `/opt/bitnami/fluentd/logs` which is a standard path for most bitnami images.

falfaro: Done.

PR #223, opened 2018-11-12
Change `--manifests` default to point to versioned github URL

### LINKED ISSUE #212: `--manifests` should default to correct version-based github URL
BKPR is now a public repo, so we can use/test the ability to fetch manifests by URL.

We should fix the URL such that `kubeprod install ...` *without* explicit `--manifests` just defaults to pointing to the right version of manifests via github URLs.

anguslees:
Make the `kubeprod install --manifests` flag default to a URL derived
from the git tag/sha used at build-time.  This should "just work" in
most situations.

Fixes #212

### DIFF
--- a/kubeprod/Makefile
+++ b/kubeprod/Makefile
@@ -1,4 +1,5 @@
-VERSION ?= dev-$(shell date +%FT%T%z)
+VERSION ?= $(shell git describe --tags --dirty)
+GIT_TAG ?= $(shell git rev-parse HEAD)
 TARGETS ?= darwin/amd64 linux/amd64 windows/amd64
 MANIFEST_DIST_DIRS = lib components platforms
 
@@ -6,7 +7,7 @@ PACKAGE = kubeprod
 
 GO = go
 GOFLAGS =
-GOBUILDFLAGS = $(GOFLAGS) -ldflags='-X main.version=$(VERSION)'
+GOBUILDFLAGS = $(GOFLAGS) -ldflags='-X main.version=$(VERSION) -X main.gitTag=$(GIT_TAG)'
 GORELEASEFLAGS = $(GOBUILDFLAGS) -tags netgo -installsuffix netgo
 GOTESTFLAGS = $(GOFLAGS) -race
 GOFMT = gofmt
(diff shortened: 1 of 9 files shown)


kubeprod/Makefile
```diff
@@ -1,12 +1,13 @@
-VERSION ?= dev-$(shell date +%FT%T%z)
+VERSION ?= $(shell git describe --tags --dirty)
```
sameersbn: Please add this to the top-level makefile for consistency.

anguslees: Oh, I missed the top-level Makefile.  I've added the new definition of VERSION & GIT_TAG, and also changed a few places where VERSION should maybe be GIT_TAG.

I think we still expect VERSION==GIT_TAG for actual releases though.  .. and given the difficulty in testing these bits of the release process, I'm ok reverting those changes (to continue just using VERSION) if you want me to do so.

kubeprod/cmd/install.go
```diff
@@ -31,8 +31,8 @@ import (
 )
 
 const (
-	FlagManifests       = "manifests"
-	DefaultManifestBase = "https://github.com/bitnami/kube-prod-runtime/manifests/"
+	FlagManifests          = "manifests"
+	defaultManifestBaseFmt = "https://github.com/bitnami/kube-prod-runtime/raw/%s/manifests/"
```
sameersbn: While testing the PR i found the right gitTag is not applied  and always appears to be set to `master` (default) instead of the one configured at build time.

anguslees: The first iteration of this PR had that issue.  Please retry with something >= 9baf12cff448a0985901156482a6e13fc8138f58

anguslees: gah, I get the same thing as you too.  I thought once I'd fixed the default shown in `--help` that that would propagate through the rest, but apparently not :(

anguslees: ok, fixed for real this time.  PTAL (the only change is `cmd/root.go:UpdateFlagDefaults()`)

PR #224, opened 2018-11-12
Collect prometheus metrics from externaldns

anguslees:


anguslees:
> We need to decide if we want this to be included in the 1.0.0 release

I vote yes, based mostly on the fact this adds a livenessprobe/readinessprobe - and therefore makes rollouts and changes more robust.

sameersbn:
I agree. I'll include it for the next RC

PR #232, opened 2018-11-15
Allow bypassing actual deployment of BKPR by the `install` command.

falfaro:
Implements a command-line flag named `--only-generate` in the `install` command that allows skipping the part that deploys BKPR to the Kubernetes cluster. The idea is to configure the underlying platform (e.g. DNS zone, security principals, etc.) and then generate the root manifest and JSON configuration files. Then, the user can add overrides and customise the root manifest, then run `kubeprod` again without the `--only-generate` command-line flag to get BKPR deployed.

### DIFF
--- a/kubeprod/cmd/install.go
+++ b/kubeprod/cmd/install.go
@@ -34,7 +34,6 @@ const (
 	FlagManifests          = "manifests"
 	defaultManifestBaseFmt = "https://github.com/bitnami/kube-prod-runtime/raw/%s/manifests/"
 	FlagOnlyGenerate       = "only-generate"
-	FlagOnlyGenerateShort  = "g"
 	FlagPlatformConfig     = "config"
 )
 
@@ -53,7 +52,7 @@ func init() {
 
 	InstallCmd.PersistentFlags().String(FlagManifests, DefaultManifestBase(), "Base URL below which to find platform manifests")
 	InstallCmd.PersistentFlags().String(FlagPlatformConfig, prodruntime.DefaultPlatformConfig, "Path for generated platform config file")
-	InstallCmd.PersistentFlags().BoolP(FlagOnlyGenerate, FlagOnlyGenerateShort, false, "Whether to just generate the root manifest and configuration files")
+	InstallCmd.PersistentFlags().Bool(FlagOnlyGenerate, false, "Stop before pushing configuration to the cluster")
 }
 
 // Common initialisation for platform install subcommands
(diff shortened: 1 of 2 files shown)


kubeprod/cmd/install.go
```diff
@@ -33,8 +33,9 @@ import (
 const (
 	FlagManifests          = "manifests"
 	defaultManifestBaseFmt = "https://github.com/bitnami/kube-prod-runtime/raw/%s/manifests/"
-
-	FlagPlatformConfig = "config"
+	FlagOnlyGenerate       = "only-generate"
+	FlagOnlyGenerateShort  = "g"
```
anguslees: I wouldn't add a short option for this - it seems too complicated an option to summarise as "`-g`".  Even more so when we don't offer anything else via a short option.

kubeprod/cmd/install.go
```diff
@@ -52,6 +53,7 @@ func init() {
 
 	InstallCmd.PersistentFlags().String(FlagManifests, DefaultManifestBase(), "Base URL below which to find platform manifests")
 	InstallCmd.PersistentFlags().String(FlagPlatformConfig, prodruntime.DefaultPlatformConfig, "Path for generated platform config file")
+	InstallCmd.PersistentFlags().BoolP(FlagOnlyGenerate, FlagOnlyGenerateShort, false, "Whether to just generate the root manifest and configuration files")
```
anguslees: Somehow we need to communicate that this also does the actual non-k8s setup (ie: it is not a read-only operation).

Maybe: `"Stop before pushing configuration to the cluster"` ?

kubeprod/pkg/installer/install.go
```diff
@@ -122,14 +122,14 @@ func (c InstallCmd) Run(out io.Writer) error {
 
 	// TODO(felipe): Conditionalize this with a command-line flag so this
 	// step is optional
-	if true {
+	if c.OnlyGenerate {
+		fmt.Println("Kubernetes cluster is ready for deployment.")
+		fmt.Println("run: `kubeprod install` without the --only-generate command-line flag.")
```
anguslees: This last line is slightly ambiguous - is it saying that's what is happening?  Does the user always have to run that next command line now? etc.

Maybe just state the simple truth: `"Exiting early because --only-generate was provided."`  (Or perhaps `"Skipping deployment because ..."`)

kubeprod/pkg/installer/install.go
```diff
@@ -122,14 +122,14 @@ func (c InstallCmd) Run(out io.Writer) error {
 
 	// TODO(felipe): Conditionalize this with a command-line flag so this
```
anguslees: This TODO can be removed!

falfaro:
PTAL

Issue #227, opened 2018-11-14
kubeprod version errors out if k8s config does not exist


sameersbn:
The `kubeprod version` command errors out when kubernetes config does not exist.
 
```bash
# kubeprod version
ERROR Error: invalid configuration: no configuration has been provided
```

The scenario can also  be reproduced with:

```bash
# KUBECONFIG=/noexist kubeprod version
ERROR Error: invalid configuration: no configuration has been provided
```

Expected output:

The command should print the correct installer version and a clear error message for the server version.

PR #234, opened 2018-11-22
Improve kubeprod error messaging on kubectl config errors

### LINKED ISSUE #82: Kubeprod returns a cyptic error message when you don't have your current-context set
Steps to reproduce:

* Unset your current-context (`kubectl config unset current-context`)
* Run kubeprod. You will get the following error:

```
ERROR Error: invalid configuration: no configuration has been provided
```

This error is quite cryptic, as it feels like the configuration you are missing is a Kubeprod configuration, rather than a missing default context.

We should provide a better error message in this case

sameersbn:
Closes #82 #227

### DIFF
--- a/kubeprod/cmd/install.go
+++ b/kubeprod/cmd/install.go
@@ -96,7 +96,7 @@ func NewInstallSubcommand(cmd *cobra.Command, platform string, config installer.
 
 	c.Config, err = clientConfig.ClientConfig()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("unable to read kubectl config: %v", err)
 	}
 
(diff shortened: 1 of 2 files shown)


kubeprod/cmd/install.go
```diff
@@ -96,7 +96,7 @@ func NewInstallSubcommand(cmd *cobra.Command, platform string, config installer.
 
 	c.Config, err = clientConfig.ClientConfig()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("Unable to read kubectl config: %v", err)
```
anguslees: golang errors should start with a lowercase.  I keep forgetting this, which is mostly why I'm making a fuss about it now ;)

sameersbn: fixed!

PR #238, opened 2018-11-23
Some troubleshooting information that belongs to quickstart for AKS.

falfaro:


docs/quickstart-aks.md
```diff
@@ -41,6 +41,12 @@ In this section, you will deploy an Azure Kubernetes Service (AKS) cluster using
   - `AZURE_AKS_CLUSTER` specifies the name of the AKS cluster.
   - `AZURE_AKS_K8S_VERSION` specifies the version of Kubernetes to use for creating the cluster. The [BKPR Kubernetes version support matrix](../README.md#kubernetes-version-support-matrix-for-bkpr-10) lists the base Kubernetes versions supported by BKPR. `az aks get-versions --location ${AZURE_REGION} -o table` lists the versions available in your region.
 
+* Get a list of your subscriptions:
```
sameersbn: I don't think we need to have this step here, we already have the command listed under the description of the `AZURE_SUBSCRIPTION_ID` shell variable.

docs/quickstart-aks.md
```diff
@@ -64,6 +70,22 @@ In this section, you will deploy an Azure Kubernetes Service (AKS) cluster using
...
+
+  even when retrying the operation multiple times, try this workaround:
+
+  ```bash
+  rm -fr ~/.azure
+  az login
+  ```
+
+  This will re-create your Azure CLI environment. Then, re-start this procedure from the beginning.
+
```
sameersbn: How about replacing this section with the following?

>  **Troubleshooting**:
>
>  If you notice the following error message while the cluster is being provisioned, it could indicate the Azure authentication token has expired.
>
>  ```
>  Operation failed with status: 'Bad Request'. Details: Service principal clientID: <SOME_CLIENT_ID> 
> not found in Active Directory tenant <SOME_TENANT_ID>, Please see https://aka.ms/acs-sp-help for more details.
>  ```
>  
>  Please remove the Azure profile directory with `rm -rf ~/.azure` and retry again by following the steps listed in [setting up the cluster](#step-1-set-up-the-cluster).

vikram-bitnami: +1. Please also think about whether you want troubleshooting at specific points in the doc, or a separate section at the end with all troubleshooting info merged.

anguslees:
This sounds like it's a bug in `kubeprod/pkg/aks/aks.go`, and I'm unhappy with such a severe workaround :(

Do we have a proper bug report for this?
Can we reproduce it easily?
Is running `az login` again sufficient to fix it (rather than deleting all of ~/.azure)?

It would be nice to just fix the issue rather than document such a blunt hammer...

falfaro:
Agreement is moving this information into a dedicated troubleshooting document. I'll cook up a new pull request.

PR #239, opened 2018-11-23
Allow overriding the pathname of JSON configuration file

falfaro:
Actually generate a JSON configuration file named according to the value of the --config command-line flag.


anguslees:
1. We should exercise this in our automated tests somehow - perhaps by always running the tests with a non-default filename.

2. What's the use-case?  I understand wanting to use a location other than pwd, but I also feel it's confusing for the user to need duplicate unintuitively-named options for each of the "user" json and jsonnet files. Worse, they have different semantics wrt overwriting, etc.   I feel like we should have a more coherent plan here before we add any further user-visible knobs.

PR #240, opened 2018-11-27
Adapt Kibana manifests to non_root container

juan131:
This PRs adapts Kibana manifests so the proper securityContext is set since Bitnami Docker image for Kibana has been moved to non-root approach.

Please **DO NOT MERGE** this PR (even if you agree with the changes) until the new image is released. I'll update this PR as soon as it's released.


juan131:
@anguslees @falfaro I just updated the image to point to the latests "non-root" version of Kibana. Feel free to test it and merge it whenever you consider it properly.

manifests/components/kibana.jsonnet
```diff
@@ -21,7 +21,7 @@ local kube = import "../lib/kube.libsonnet";
 local kubecfg = import "kubecfg.libsonnet";
 local utils = import "../lib/utils.libsonnet";
 
-local KIBANA_IMAGE = "bitnami/kibana:5.6.12-r15";
+local KIBANA_IMAGE = "bitnami/kibana:5.6.13-r20";
```
falfaro: I think the safest is to keep Elasticsearch and Kibana versions synchronized. So I would bump Elasticsearch to 5.6.13 too.

anguslees: Indeed.  I believe Kibana logs a warning if the versions don't match (even if the mismatch is minor and safe, like this example presumably is).

juan131: Bumping ElasticSearch image too so it's synchronized with Kibana

falfaro:
You will have to update this PR after https://github.com/bitnami/kube-prod-runtime/pull/280

juan131:
@falfaro I rebased from master and updated the images in the `image.json` file

falfaro:
I'll test it and take care of merging it. Thanks!

PR #244, opened 2018-11-29
Upload release manifests to S3 bucket

### LINKED ISSUE #236: Server version reported as `dev-untagged` with versioned github URL
In #223 we changed default manifest path to use the default github url making the `--manifest` argument optional. However when this is used, the manifests version is reported as `dev-untagged`, since the VERSION file is only present in the release tarballs.

f.e. I used the `v1.0.0-rc5` to deploy BKPR and I get the following version reported back.

```bash
kubeprod version
Installer version: v1.0.0-rc5
Server manifests version: dev-untagged
```

sameersbn:
Changes:
- [x] Publish manifests to S3 bucket
- [x] Default manifest base url set to https://releases.kubeprod.io/files (previously github.com), configurable at build-time.
- [x] `--manifests` flag is optional only for releases
- [x] Removed unused GIT_TAG related code
- [x] availability of https://releases.kubeprod.io endpoint
 
Closes #236

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -492,7 +492,7 @@ gcloud container clusters create ${clusterName} \
 
                             sh "make dist VERSION=${TAG_NAME} GIT_TAG=${TAG_NAME}"
 
-                            withCredentials([usernamePassword(credentialsId: 'github-bitnami-bot', passwordVariable: 'GITHUB_TOKEN', usernameVariable: '')]) {
+                            withCredentials([usernamePassword(credentialsId: 'github-bitnami-bot', passwordVariable: 'GITHUB_TOKEN', usernameVariable: ''),]) {
                                 sh "make publish VERSION=${TAG_NAME} GIT_TAG=${TAG_NAME}"
                             }
                         }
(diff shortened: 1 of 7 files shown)


sameersbn:
Sorry if you find the PR has a few "extra" changes. Follow the commit history for targeted changes

kubeprod/cmd/install.go
```diff
@@ -30,11 +31,13 @@ import (
 	"github.com/bitnami/kube-prod-runtime/kubeprod/tools"
 )
 
+var ReleaseFilesBaseUrl = "https://releases.kubeprod.io/files"
```
sameersbn: This end-point is not online yet, should be up today.

Jenkinsfile
```diff
@@ -102,7 +102,7 @@ def runIntegrationTest(String description, String kubeprodArgs, String ginkgoArg
 
                 sh "kubectl --namespace kube-system get po,deploy,svc,ing"
 
-                sh "./bin/kubeprod -v=1 install --config=kubeprod-autogen.json ${kubeprodArgs}"
+                sh "./bin/kubeprod -v=1 install --manifests=manifests --config=kubeprod-autogen.json ${kubeprodArgs}"
```
falfaro: Maybe we should use a different name for --config to exercise generating the JSON configuration file under a different name. Something like --config=kubeprod.json?

sameersbn: makes sense. I'll do it in a different PR

PR #252, opened 2018-12-03
Upgrade + actually scrape node-exporter

### LINKED ISSUE #248: [node-exporter]:  Add scraping annotations  and upgrade image to latest
Add scraping annotations:
```
               annotations+: {
                 "prometheus.io/path": "/metrics",
                 "prometheus.io/scrape": "true",
               }
```
upgrade image to latest:
```                   
 image: "bitnami/node-exporter:v0.16.0",
```

anguslees:
Add missing annotations and actually scrape the node exporter.

Update node-exporter 0.15.2 -> 0.17.0
(Note 0.16.0 included "major breaking changes to metric names."  The
new names follow current naming conventions (so it's a good thing).)

Adds an integration test that catches the original issue.

Also: add missing (less critical) scrape annotations on fluentd-es, noticed while writing the integration test.

Fixes #248


manifests/components/prometheus.jsonnet
```diff
@@ -394,7 +401,7 @@ local get_cm_web_hook_url = function(port, path) (
             }],
             containers_+: {
               default: kube.Container("node-exporter") {
-                image: "prom/node-exporter:v0.15.2",
+                image: "bitnami/node-exporter:v0.17.0-r1",
```
sameersbn: There appear to be a large number of breaking changes since 0.15.x release (https://github.com/prometheus/node_exporter/blob/master/CHANGELOG.md). Curious to know if this upgrade is safe for BKPR?

dbarranco: In our case when we upgraded from `0.14.X` to `0.16.0` we got to do some changes, but related with Grafana dashboards and the naming of the metrics. In this case it shouldn't be a problem

jjo: indeed, nothing functionality-wise but only metrics naming fixes

sameersbn: Thanks for the clarification

anguslees: jjo's earlier call for more testing was absolutely correct.  The image was "...:0.17.0-r1" (no preceding "v" on the tag), and this PR was broken in its original form.  In some ways this is a good thing, since we (as humans) should not have to scrutinise the image name/tags for basic validity and I'm kind of glad that we've trained ourselves not to perform that level of study.

On the other hand, that means our automated tests need to actually exist, and we need to learn to demand them :stuck_out_tongue: 

I have added an integration test to this PR, which I can confirm catches the original missing annotations, and (by side effect) the above incorrect node-exporter image.  It also prompted me to realise that our fluentd-es jobs were also missing annotations, so I've fixed that too.  Win.

sameersbn:
integration tests lgtm!

tests/monitoring_test.go
```diff
@@ -81,6 +84,58 @@ func countAlerts(alerts []alert) int {
...
+			fmt.Fprintf(GinkgoWriter, "%s{", s["__name__"])
+			for k, v := range s {
+				if k != "__name__" {
+					fmt.Fprintf(GinkgoWriter, "%s=%q,", k, v)
+				}
+			}
+			fmt.Fprintf(GinkgoWriter, "}\n")
+		}
+
+		Expect(series).To(match)
```
sameersbn: Prometheus may not be ready when we launch the tests in the CI. We need to have some retries in this test.

falfaro:
This fixes #248.

tests/monitoring_test.go
sameersbn: The `kubectl rollout` command can only report rollout of Deployments. Prometheus and Elasticsearch use Statefulsets. I couldn't find a way to make it work for statefulsets.

sameersbn: When I tried using `kubectl rollout` with a sts, I remember it would always report that it was not ready despite being actually rolled out.

sameersbn: Added https://github.com/bitnami/kube-prod-runtime/pull/257 to support STS

PR #256, opened 2018-12-04
Upgrade to NGINX ingress controller 0.21.

### LINKED ISSUE #249: upgrade nginx-ingress-controller to 0.21
https://github.com/kubernetes/ingress-nginx/releases/tag/nginx-0.21.0 fixes
http/2 security issues, we should upgrade as soon as we have
https://github.com/bitnami/bitnami-docker-nginx-ingress-controller/issues/1 fixed.

Note that 0.21 also deprecates the need for explicit default-http-backend,
we should also remove it from the manifests (FYI we already did so in our
clusters).

falfaro:
This PR upgrades NGINX ingress controller to version 0.21 from version 0.19, and also removes the default-http-backend which is not required since 0.20.

Fixes #249.

### DIFF
--- a/manifests/components/nginx-ingress.jsonnet
+++ b/manifests/components/nginx-ingress.jsonnet
@@ -19,7 +19,7 @@
 
 local kube = import "../lib/kube.libsonnet";
 
-local NGNIX_INGRESS_IMAGE = "bitnami/nginx-ingress-controller:0.19.0-r8";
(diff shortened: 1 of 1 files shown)


jjo:
BTW suggest also adding a note and/or overloading PR description re: default-http-backend removal

falfaro:
@jjo added a comment to the PR (and eventually the commit) about removal of the default HTTP backend.

PR #257, opened 2018-12-05
ci: add support for Statefulset in `waitForRollout`

sameersbn:


Jenkinsfile
```diff
@@ -35,10 +35,10 @@ def waitForRollout(String namespace, int minutes) {
             timeout(time: minutes, unit: 'MINUTES') {
                 sh """
 set +x
-for deploy in \$(kubectl --namespace ${namespace} get deploy --output name)
+for deploy in \$(kubectl --namespace ${namespace} get deploy,sts --output name)
 do
   echo -n "\nWaiting for rollout of \${deploy} in ${namespace} namespace"
-  while ! \$(kubectl --namespace ${namespace} rollout status \${deploy} --watch=false | grep -q "successfully rolled out")
+  while ! \$(kubectl --namespace ${namespace} rollout status \${deploy} --watch=false | grep -q -E 'successfully rolled out|pods have been updated|rolling update complete')
```
anguslees: Does this work?

```console
# In this cluster, statefulset/prometheus is already rolled out
$ kubectl -n kubeprod rollout status statefulset/prometheus
Waiting for partitioned roll out to finish: 0 out of 1 new pods have been updated...
$ kubectl version --client --short
Client Version: v1.12.2
```

I think searching for `pods have been updated` will match the above string and exit too early.  Otoh, I think kubectl itself will not correctly detect the rollout of a statefulset(?).

sameersbn: There is no one single status message that can be used to consider the sts rolled out, it's just a poor attempt  at evaluating that the rollout has been completed and it works :slightly_smiling_face: 

p.s. Recently I updated the client on the CI server to 1.12.x

anguslees: Fwiw, the relevant code (and output strings) are here: https://github.com/kubernetes/kubernetes/blob/0214031fcf939acab415779764e2a3dfcaba55c2/pkg/kubectl/rollout_status.go#L121

I think it's stuck in my case because `sts.Status.UpdatedReplicas` is not in the reply object (so always defaults to 0)

sameersbn: > I think searching for pods have been updated will match the above string and exit too early

The issue is that it never changes to `statefulset rolling update complete`, so...

sameersbn: > I think it's stuck in my case

I see the same for `statefulset.apps/prometheus`

```bash
# kubectl --context ${kubeContext} -n kubeprod rollout status statefulset.apps/prometheus --watch=false
Waiting for partitioned roll out to finish: 0 out of 1 new pods have been updated...
```

manifests/components/prometheus.jsonnet
```diff
@@ -202,6 +202,8 @@ local get_cm_web_hook_url = function(port, path) (
             storage: "%dMi" % $.storage,
           },
         },
+        replicas: 1,
+        updateStrategy: { type: "RollingUpdate" },
```
sameersbn: @anguslees this change gets the rollout status to be reported cleanly. Do you see any issues with this update?

manifests/lib/kube.libsonnet
```diff
@@ -384,9 +384,6 @@
 
       updateStrategy: {
         type: "RollingUpdate",
-        rollingUpdate: {
-          partition: 0,
-        },
```
sameersbn: Fixes the rollout issue for statefulsets. From https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#partitions it appears that it does not make sense to have it defined by default.

sameersbn: /cc @anguslees

anguslees: Nice find!

Oddly, the default for that value is meant to be 0, but it's actually a `*int32`, so `nil` is different to `0`.  So much for the docs :stuck_out_tongue:

PR #258, opened 2018-12-05
Better explanation on failure to set IAM policy for GKE.

### LINKED ISSUE #230: Failed to set IAM policy
`kubeprod install` on GKE can fail when the authenticated user lacks certain privileges on GCE/GKE. In this bug I have observed the following error message:

```
INFO  Installing platform gke
INFO  Using existing Google DNS zone "gke.felipe-alfaro.com"
INFO  Created service account bkpr-edns-gke-felipe-alfaro-co@bitnamigcetest2.iam.gserviceaccount.com
ERROR Error: failed to set IAM policy on project "XXX": googleapi: Error 403: The caller does not have permission, forbidden
```

which seems to be caused because the user running `kubeprod` does not have privileges to grant the `roles/dns.admin` role to the service account used to manage the DNS zone.

The offending code:

```
                        policy, err = crmService.Projects.
                                SetIamPolicy(conf.ExternalDNS.Project, &crm.SetIamPolicyRequest{
                                        Policy: policy,
                                }).
                                Context(ctx).
                                Do()
                        if err != nil {
                                // FIXME: retry on etag conflict
                                return fmt.Errorf("failed to set IAM policy on project %q: %v", conf.ExternalDNS.Project, err)
                        }
```

It would be nice if `kubeprod` was able to report a more accurate description of the problem and possible workarounds. For example, workarounds can be documented by pointing the user to some knowledge base article or documentation page for `kubeprod`.

falfaro:
Fixes #230

### DIFF
--- a/kubeprod/pkg/gke/platform.go
+++ b/kubeprod/pkg/gke/platform.go
@@ -270,7 +270,8 @@ func (conf *GKEConfig) Generate(ctx context.Context) error {
 				Context(ctx).
 				Do()
 			if err != nil {
-				return fmt.Errorf("failed to get IAM policy for project %q: %v", conf.ExternalDNS.Project, err)
+				return fmt.Errorf("failed to get IAM policy for project %q: %v",
+					conf.ExternalDNS.Project, err)
 			}
 
 			addIamBinding(policy, "roles/dns.admin", "serviceAccount:"+sa.Email)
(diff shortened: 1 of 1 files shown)


kubeprod/pkg/gke/platform.go
```diff
@@ -283,7 +284,9 @@ func (conf *GKEConfig) Generate(ctx context.Context) error {
 				Do()
 			if err != nil {
 				// FIXME: retry on etag conflict
-				return fmt.Errorf("failed to set IAM policy on project %q: %v", conf.ExternalDNS.Project, err)
+				return fmt.Errorf("failed to set IAM policy on project %q: user lacks "+
```
anguslees: I don't think we can state this with authority without at least checking `err` for something that indicates "permission denied".  (Eg, consider how confusing this error message would be if `err` was "connection refused")

I think we just have to add some weasel-words to suggest it's just a guess:
`"failed to set IAM policy on project %q (possibly caused by not having privileges to grant roles/dns.admin role to service account %s): %v"`

PR #260, opened 2018-12-05
Fixes to Grafana.

falfaro:
- Rename ConfigMap to `grafana-prometheus-datasource`
- Fix ConfigMap for Grafana-Prometheus integration
- Mount the PVC under /opt/grafana/data

### DIFF
--- a/manifests/components/grafana.jsonnet
+++ b/manifests/components/grafana.jsonnet
@@ -23,6 +23,7 @@ local utils = import "../lib/utils.libsonnet";
 
 local GRAFANA_IMAGE = "bitnami/grafana:5.3.4-r6";
 local GRAFANA_DATASOURCES_CONFIG = "/opt/bitnami/grafana/conf/provisioning/datasources";
+local GRAFANA_DATA_MOUNTPOINT = "/opt/bitnami/grafana/data";
 
 // TODO: add blackbox-exporter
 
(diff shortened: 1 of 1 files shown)


dbarranco:
LGTM, tested that dashboards are persisted after kicking the Grafana pod.

manifests/components/grafana.jsonnet
```diff
@@ -68,15 +69,14 @@ local GRAFANA_DATASOURCES_CONFIG = "/opt/bitnami/grafana/conf/provisioning/datas
...
     datasources:: {
       // Built-in datasource for BKPR's Prometheus
       "BKPR Prometheus": {
         type: "prometheus",
-        access: "server",
-        InsecureSkipVerify: "true",
-        orgId: 1,
+        access: "proxy",
+        isDefault: true,
         url: "http://%s:9090/" % $.prometheus.host,
```
anguslees: I know it isn't part of this change, but: `url: $.prometheus.http_url` ?

anguslees: (I'm also ok with merging this as-is)

dbarranco: I think `prometheus.http_url` equals to the ingress URL of Prometheus, and now that Prometheus is configured behind the oAuth proxy it can affect to the queries, that's why is configured as the internal URL (`prometheus.kubeprod.svc`)

falfaro: @gus is correct and $.prometheus.http_url generates `url: http://prometheus.kubeprod.svc:9090/`

PR #264, opened 2018-12-05
Minor update in GKE quick start

migmartri:


falfaro:
LGTM!

anguslees:
The original description for that value is:
```markdown
 `GCLOUD_USER` specifies the email address used in requests to Let's Encrypt.
```

I agree it has the same _default_ value in the instructions, and the name "GCLOUD_USER" sounds like it should always have that value.  This *description* however sounds like it should be a personal/antispam email address, which may not be the same as your GCP username.

TL;DR: For a quickstart, I think it's fine to reuse the GCP auth email for the letsencrypt contact.  I suggest we change the description for GCLOUD_USER to be "the email address used to authenticate to Google Cloud Platform."

sameersbn:
Thanks for the contribution @migmartri. Could you also update the description of the `GCLOUD_USER` field as suggested by @anguslees ?

sameersbn:
Thanks for the contribution @migmartri. Like @anguslees pointed out the two email addresses are not necessarily always the same. In the current iteration of the doc, `GCLOUD_USER` configures the Let's encrypt contact email address that's used in lets encrypt requests (i agree it's a poorly named variable :slightly_smiling_face:). Whereas while creating the clusterrolebinding, as Gus pointed out, we have to specify the email address used to authenticate to Google Cloud Platform.

Could you also update the description of the `GCLOUD_USER` field as suggested by Gus and also update the default value of `GCLOUD_USER` to `$(gcloud info --format='value(config.account)')`

migmartri:
Just updated it, but I am a little bit confused TBH. 

The `email` flag provided during installation is the email address for lets encrypt then? https://github.com/bitnami/kube-prod-runtime/pull/264/files#diff-d9eaa84d62d91998f096c8ff57ace5e2L117

>  and also update the default value of GCLOUD_USER to $(gcloud info --format='value(config.account)')

This is already the case isn't it?

Thanks!

sameersbn:
> The `email` flag provided during installation is the email address for lets encrypt then?

This is right. We're using the email address only for the lets encrypt setup. But I would not say that it's future use would be limited to just that. We would be using it in any setup that requires the cluster administrators email.

> This is already the case isn't it?

Oh yes, that's right!

migmartri:
Thanks for confirming it @sameersbn!

PR #280, opened 2018-12-18
Keep all image versions in a single JSON file.

falfaro:
This should make release engineering and automated updates of
image versions easier.

Also, bump versions for:

* cert-manager
* external-dns

### DIFF
--- a/manifests/components/cert-manager.jsonnet
+++ b/manifests/components/cert-manager.jsonnet
@@ -18,6 +18,7 @@
  */
 
 local kube = import "../lib/kube.libsonnet";
+local CERT_MANAGER_IMAGE = "bitnami/cert-manager:%s" % (import "versions.json")["cert-manager"];
 
 {
   p:: "",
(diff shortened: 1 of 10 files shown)


falfaro:
Yes, it is similar to your previous take at this and we need something in place in order to be able to automate part of the release engineering when new assets are released (e.g. due to security issues).

Do you think that it is not easy to parse and manipulate JSON? There are libraries for Go and Python and Java. You can even manipulate JSON on the command-line using `jq`.

jjo:
@falfaro sure re `jq` and friends, tho I was thinking on even simpler
ways (e.g. being able to do so via `sed` and friends).
Nevertheless, this is better than existing, LGTM'ing.

arapulido:
Why we are not doing it for all the images at once?

falfaro:
I am going to do this for all images, and keep the versions unchanged (0-diff).

falfaro:
Done. PTAL.

falfaro:
The logic is so simple that I think no helper functions are needed, but if necessary I could do that.

PR #294, opened 2019-01-02
ci: use preemptible vm's for gke jobs

sameersbn:
Preemptible VM's last a maximum of 24 hours and are priced lower than
standard Compute Engine VMs. As the CI jobs create short-lived clusters
we can use preemptible vm's for it's use and signiificantly reduce ci
costs (upto 80%).

ref: https://cloud.google.com/preemptible-vms/


arapulido:
This is great, thanks @sameersbn !

PR #298, opened 2019-01-04
docs: rename quickstart env var `GCLOUD_REGION` to `GCLOUD_ZONE`

### LINKED ISSUE #285: BKPR on GKE Quickstart
I'm following the BKPR quickstart on GKE found here:
https://github.com/bitnami/kube-prod-runtime/blob/master/docs/quickstart-gke.md

The page describes setting the GCLOUD_REGION variable as follows:

> GCLOUD_REGION specifies the Google Cloud region. gcloud compute regions list lists the available Google Cloud regions.

However the example lists a zone as the value "us-east1-d" This is further confused by the section below, which also references regions and suggests using ```gcloud compute regions list``` to get a list of possible values.

In my run I set the GCLOUD_REGION=us-west2-c and my cluster appears to have deployed correctly, at least in this regard. If you could clarify what the intended value is I can prepare a PR with some improved text for this quickstart for your review?

sameersbn:
The user should specify the GCE zone (not region). Corrected this
mistake and also renamed the variable to `GCLOUD_ZONE`.

Fixes #285

### DIFF
--- a/docs/quickstart-gke.md
+++ b/docs/quickstart-gke.md
@@ -25,7 +25,7 @@ In this section, you will deploy a Google Kubernetes Engine (GKE) cluster using 
   export BKPR_DNS_ZONE="my-domain.com"
   export GCLOUD_USER="$(gcloud info --format='value(config.account)')"
   export GCLOUD_PROJECT="my-gce-project"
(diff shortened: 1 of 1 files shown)


nomisbeme:
Looks good. Thanks!

PR #305, opened 2019-01-07
docs: specify dns and g suite requirements in the quickstart

### LINKED ISSUE #279: Docs for BKPR on GKE could use improvement
This is based on my experience with the GKE quickstart  (https://github.com/bitnami/kube-prod-runtime/blob/master/docs/quickstart-gke.md) .  The **Prerequisites** section should have a subsection which should include "**G Suite account and DNS setup**" or something to that effect. 

G Suite is mentioned later in the README. However, it turned out to be a big deal to setup a new domain name via a registrar and link that to a newly created "G Suite" (formerly known as G Apps) account. This matters later during the BKPR cluster setup, and relates to the domain specified by the  "GCLOUD_AUTHZ_DOMAIN".  

Furthermore, after a new G Suite account is created, the Kubernetes API and the Gcloud DNS API need to be enabled in GCP for the newly targeted GKE project. Those are additional "pre-requisites" needed before you can run the kubeprod install gke  command successfully. 

I showed these to Gus (and others at the Bitnami booth at KubeCon) while attempting the Gke quickstart, and ran into these issues.  So hopefully with enhanced docs, especially the pre-reqs for new Google G suite account and DNS setup, the improvements will help other newcomers.

sameersbn:
A DNS domain and a G Suite account (in case of GKE) are required to successfully install BKPR.

Closes #279


docs/quickstart-gke.md
```diff
@@ -13,12 +13,30 @@ This document walks you through setting up a Google Kubernetes Engine (GKE) clus
 * [kubecfg](https://github.com/ksonnet/kubecfg/releases)
 * [jq](https://stedolan.github.io/jq/)
 
+### DNS and G Suite requirements
+
+In addition to the requirements listed above, a domain name is also required for setting up Ingress endpoints to services running in the cluster. The specified domain name can be a top-level domain (TLD) or a subdomain. In either case you have to manually [set up the glue records](#configure-domain-registration-records)  to delegate DNS resolution queries to a Google Cloud DNS zone created and managed by BKPR.
+
+BKPR on GKE uses AuthZ domain based authorization, specified in `GCLOUD_AUTHZ_DOMAIN`, for verifying users as they login to the Elasticsearch, Kibana and Grafana dashboards. As such you need a [G Suite](https://gsuite.google.co.in/intl/en_in/) account configured set up for your authorization domain to enable users to login with their G Suite user accounts.
+
```
sameersbn: @falfaro @anguslees Do you think we should note here that `GCLOUD_AUTHZ_DOMAIN` can optionally be configured to `*` to allow users to login with their @gmail.com email address?

sameersbn: Just realized it's mentioned when `GCLOUD_AUTHZ_DOMAIN` is described later.

docs/quickstart-aks.md
```diff
@@ -13,6 +13,10 @@ This document walks you through setting up an Azure Kubernetes Service (AKS) clu
 * [kubecfg](https://github.com/ksonnet/kubecfg/releases)
 * [jq](https://stedolan.github.io/jq/)
 
+### DNS requirements
+
+In addition to the requirements listed above, a domain name is also required for setting up Ingress endpoints to services running in the cluster. The specified domain name can be a top-level domain (TLD) or a subdomain. In either case you have to manually [set up the NS records](#configure-domain-registration-records) for the specified TLD or subdomain so as to delegate DNS resolution queries to a Azure DNS zone created and managed by BKPR.
```
anguslees:
```suggestion
In addition to the requirements listed above, a domain name is also required for setting up Ingress endpoints to services running in the cluster. The specified domain name can be a top-level domain (TLD) or a subdomain. In either case you have to manually [set up the NS records](#configure-domain-registration-records) for the specified TLD or subdomain so as to delegate DNS resolution queries to an Azure DNS zone created and managed by BKPR.  This is required in order to generate valid TLS certificates.
```

docs/quickstart-gke.md
```diff
@@ -7,18 +7,39 @@ This document walks you through setting up a Google Kubernetes Engine (GKE) clus
...
 * [Kubernetes CLI](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
 * [BKPR installer](install.md)
 * [kubecfg](https://github.com/ksonnet/kubecfg/releases)
 * [jq](https://stedolan.github.io/jq/)
 
+### DNS and G Suite requirements
+
+In addition to the requirements listed above, a domain name is also required for setting up Ingress endpoints to services running in the cluster. The specified domain name can be a top-level domain (TLD) or a subdomain. In either case you have to manually [set up the NS records](#configure-domain-registration-records) for the specified TLD or subdomain so as to delegate DNS resolution queries to a Google Cloud DNS zone created and managed by BKPR.
+
+BKPR on GKE uses AuthZ domain based authorization, specified in `GCLOUD_AUTHZ_DOMAIN`, for verifying users as they login to the Elasticsearch, Kibana and Grafana dashboards. As such you need a [G Suite](https://gsuite.google.co.in/intl/en_in/) account configured set up for your authorization domain to enable users to login with their G Suite user accounts.
```
anguslees:
```suggestion
BKPR on GKE uses domain-based authorization, specified in `GCLOUD_AUTHZ_DOMAIN`, for verifying users as they login to the Elasticsearch, Kibana and Grafana dashboards. As such you need a [G Suite](https://gsuite.google.com/) account configured set up for your authorization domain to enable users to login with their G Suite user accounts.
```

docs/quickstart-gke.md
anguslees: We have a duty of care here and I think it is irresponsible for us to suggest that a user allow _all_ @gmail.com users access to their site - particularly when it won't be obvious that they've just done this.

I know we don't like requiring a Gsuite domain, but we need to fix that by exposing the oauth2_proxy email-address whitelist rather than suggesting (anywhere!) that they effectively just turn off authz.

If we _really_ want (and I don't think this is a good idea), we could implement an `--insecure` mode that doesn't use TLS, doesn't require real hostnames (no letsencrypt), and skips all of oauth2_proxy.  We still need hostnames in _some_ form if we continue to choose the service based on HTTP Host headers, but we could even drop that (to use multiple addresses or multiple ports) if we really want to drop all environmental requirements.

sameersbn:
Updated! PTAL @anguslees

docs/quickstart-gke.md
sameersbn: I agree! I have updated the PR to remove the suggestion for specifying `GCLOUD_AUTHZ_DOMAIN="*"` to (kinda) bypass the g suite requirement

sameersbn:
p.s. also updated the troubleshooting guide to document issues faced in #279

sameersbn:
@conradf7 would be great if you'd take a look at these changes

PR #343, opened 2019-01-29
ExternalDNS: Track provenance in non-conflicting TXT name

anguslees:
Previously ExternalDNS stored a TXT record alongside each managed DNS
record, to avoid stomping on DNS entries that were created
elsewhere (good).

These TXT records conflicted with CNAME records, since they had the
same name (bad).

Configure ExternalDNS to store the TXT record in a different name.
Now the TXT metadata record for `foo` will be stored as
`_externaldns.foo`.


sameersbn:
@falfaro you'd have to rebase this PR to fix the CI issues

PR #345, opened 2019-01-29
ci: automatically use latest k8s PATCH versions for AKS test clusters

sameersbn:
For a given k8s series (MAJOR.MINOR), We now query the AKS API and use the
most recent PATCH release available of the specified version series.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -256,7 +256,7 @@ spec:
     // See:
     //  az aks get-versions -l centralus
     //    --query 'sort(orchestrators[?orchestratorType==`Kubernetes`].orchestratorVersion)'
-    def aksKversions = ["1.10.12", "1.11.6"]
+    def aksKversions = ["1.10", "1.11"]
     for (x in aksKversions) {
         def kversion = x  // local bind required because closures
         def platform = "aks-" + kversion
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -289,6 +289,7 @@ az account set -s $AZURE_SUBSCRIPTION_ID
                                             // a) avoid this leak b) avoid having to give the
                                             // "outer" principal (above) the power to create
                                             // new service principals.
+                                            def kversion_full = sh(returnStdout: true, script: "az aks get-versions --location ${location} --query \"orchestrators[?contains(orchestratorVersion,'${kversion}')].orchestratorVersion\" -o tsv | tail -n1 | tr -d '\n'")
```
falfaro: This could fail because the list of K8s versions is not properly sorted. Instead use:

```
az aks get-versions --location ${location} --query \"orchestrators[?contains(orchestratorVersion,'${kversion}')].orchestratorVersion\" -o tsv | sort -t. -k1n -k2n -k3n | tail -n1 | tr -d '\n'
```

sameersbn: unfortunately that does not work, fe. for input file named `test` with the following content

```
1.10.12
1.10.9
1.10.3
```

```
# cat test | sort -t. -k1n -k2n -k3n
1.10.12
1.10.3
1.10.9
```

while this is a proper numerical sort (can also be achieved with `sort -n`),  it's not appropriate to sort version strings. Originally i used `sort -V` which is custom built to sort versions. But unfortunately the busybox version of `sort` does not implement the `-V` flag, so I just retired with the assumption that the aks api would return in the desired sort order.

I spend some time to see if I can get the proper `sort` utility installed and use the `sort -V` instead.

sameersbn: @falfaro Turns out the `sort` command from the parent container has the `coreutils` version of `sort` that implements the convenient `-V` flag. I updated the PR to perform this sorting outside the `az` container. can you PTAL. Thanks!

PR #346, opened 2019-01-29
OAuth2 proxy support for EKS in kubeprod

falfaro:
Add missing support to kubeprod for EKS: integration between OAuth2 proxy and AWS Cognito.


kubeprod/pkg/eks/platform.go
```diff
@@ -305,6 +426,9 @@ func (conf *Config) Generate(ctx context.Context) error {
 			return err
 		}
 		conf.ContactEmail = email
+		if conf.ContactEmail == "" {
```
anguslees: Hrm - we don't do this check in other platforms.  Is it required? (does cert-manager fail to function with an empty-string email address?) 
If it is required, we should probably mark the flag as required instead of this check (and do the same across the other platforms).

kubeprod/pkg/eks/platform.go
```diff
@@ -295,6 +299,123 @@ func (conf *Config) setUpExternalDNS() error {
...
+	return ok
+}
+
+// Configuration for integration between OAuth2 Proxy and AWS Cognito.
+func (conf *Config) setUpOAuth2Proxy() error {
+	if conf.OauthProxy.ClientID == "" || conf.OauthProxy.ClientSecret == "" {
+		log.Info("Setting up configuration for OAuth2 Proxy")
+
+		// Extract the AWS region name/ID from the Cognito user pool ID
+		conf.OauthProxy.AWSRegion = strings.Split(conf.OauthProxy.AWSUserPoolID, "_")[0]
```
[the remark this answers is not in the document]
falfaro: Fixed. I have changed the code that instantiates the AWS session to load the SDK configuration, which includes the default AWS region.

kubeprod/pkg/eks/platform.go
```diff
@@ -295,6 +299,123 @@ func (conf *Config) setUpExternalDNS() error {
...
+	}
+
+	if conf.OauthProxy.CookieSecret == "" {
+		// Generate a 32-byte random for securing cookies
+		key := make([]byte, 16)
+		_, err := rand.Read(key)
+		if err != nil {
+			return err
+		}
+		conf.OauthProxy.CookieSecret = fmt.Sprintf("%x", key)
```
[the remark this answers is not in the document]
falfaro: Fixed.

kubeprod/pkg/eks/platform.go
falfaro: Yes, cert-manager will fail because the letsencrypt-prod clusterissuer will get upset if an e-mail is not provided.

falfaro: I'll disable the requirement for now, and will fix this in a separate PR (to make the e-mail mandatory).

falfaro:
@anguslees PTAL

PR #368, opened 2019-02-01
Testing for EKS support in kubeprod

falfaro:


anguslees:
>  I am blindly adding manifests/tests/eks.jsonnet [..] but I believe this file is unused?

It is used by `make -C manifests validate`, run (early) by Jenkinsfile.

Jenkinsfile
```diff
@@ -488,6 +499,91 @@ gcloud container clusters create ${clusterName} \
...
+                                            writeFile([file: "${clusterName}-autogen.json", text: """
+{
+  "dnsZone": "${dnsZone}",
+  "contactEmail": "${adminEmail}",
+  "externalDns": {
+    "aws_access_key_id": "${AWS_ACCESS_KEY_ID}",
+    "aws_secret_access_key": "${AWS_SECRET_ACCESS_KEY}"
+  },
+  "oauthProxy": {
+    "client_id": "${fixme}",
```
anguslees: These need appropriate values

Jenkinsfile
```diff
@@ -488,6 +499,91 @@ gcloud container clusters create ${clusterName} \
...
+                                            writeFile([file: 'kubeprod-manifest.jsonnet', text: """
+(import "manifests/platforms/eks.jsonnet") {
+  config:: import "${clusterName}-autogen.json",
+  letsencrypt_environment: "staging",
+  prometheus+: import "tests/testdata/prometheus-crashloop-alerts.jsonnet",
+}
+"""
+                                            ])
+                                        }
+                                    }{
```
anguslees: Syntax error here somewhere:
```console
org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
WorkflowScript: 567: unexpected token: { @ line 567, column 38.
                                       }{
                                        ^
```

Jenkinsfile
```diff
@@ -488,6 +499,91 @@ gcloud container clusters create ${clusterName} \
...
+                                        container('aws') {
+                                            def output = sh(returnStdout: true, script: """
+id=$(aws route53 list-hosted-zones --query 'HostedZones[?Name==`bkpr.run.`].Id' --output text)
+aws route53 get-hosted-zone --id $id --query DelegationSet.NameServers
+""")
+                                            def nameServers = readJson(text: output)
+                                            insertGlueRecords(dnsPrefix, nameServers, "60", parentZone, parentZoneResourceGroup)
+                                        }
+                                    }
+                                }
```
anguslees: oops, looks like I forgot to add cleanup code in a `finally` unwind about here.  We should call `eksctl delete cluster ...`

Jenkinsfile
```diff
@@ -488,6 +499,91 @@ gcloud container clusters create ${clusterName} \
...
     }
 
+    def eksVersions = ["1.10", "1.11"]
+    for (x in eksVersions) {
+        def kversion = x // local bind required because closures
+        def platform = "eks-${kversion}"
+        platforms[platform] = {
+            stage(platform) {
+                node(label) {
+                    withCredentials([[$class: 'AmazonWebServicesCredentialsBinding', credentialsId: 'aws-kubeprod-jenkins']]) {
```
anguslees: This jenkins secret still needs to be created.

Jenkinsfile
falfaro: Done.

Jenkinsfile
falfaro: Fixed.

Jenkinsfile
falfaro: Fixed.

Jenkinsfile
falfaro: Added to Jenkins (and changed the secret ID in Jenkins to a more meaningful name)

falfaro:
@anguslees PTAL

Jenkinsfile
```diff
@@ -23,6 +24,8 @@ def withGo(Closure body) {
             "GOPATH+WS=${env.WORKSPACE}",
             "PATH+GOBIN=${env.WORKSPACE}/bin",
             "HOME=${env.WORKSPACE}",
+            // https://docs.aws.amazon.com/eks/latest/userguide/install-aws-iam-authenticator.html
+            "PATH+AWSIAMAUTHENTICATOR=${tool 'aws-iam-authenticator'}"
```
sameersbn: I would suggest performing the tool installation only where it will be used. By defining it here, we're installing it every time the `withGo` function is called.

Jenkinsfile
```diff
@@ -490,6 +506,107 @@ az aks create                      \
...
+            stage(platform) {
+                node(label) {
+                    withCredentials([[
+                        $class: 'AmazonWebServicesCredentialsBinding',
+                        credentialsId: 'aws-eks-kubeprod-jenkins',
+                        accessKeyVariable: 'AWS_ACCESS_KEY_ID',
+                        secretKeyVariable: 'AWS_SECRET_ACCESS_KEY',
+                    ]]) {
+                        withEnv([
+                            "AWS_DEFAULT_REGION=eu-central-1",
```
sameersbn: FWIW, you can add this as an environment variable for the aws-cli container in the PodTemplate

Jenkinsfile
```diff
@@ -155,6 +162,16 @@ spec:
     tty: true
     command:
     - 'cat'
+  - name: 'eksctl'
+    image: weaveworks/eksctl
```
sameersbn: We should be specifying the docker image tag

Jenkinsfile
```diff
@@ -155,6 +162,16 @@ spec:
...
+    tty: true
+    command:
+    - 'cat'
+  - name: 'aws'
+    image: 'mesosphere/aws-cli:latest'
```
sameersbn: Avoid using `latest` tag

Jenkinsfile
```diff
@@ -490,6 +506,107 @@ az aks create                      \
             }
         }
     }
+*/
+
+    //def eksVersions = ["1.10", "1.11"]
+    def eksVersions = ["1.11"]
```
sameersbn: reminder to enable tests for other supported k8s versions

Jenkinsfile
```diff
@@ -490,6 +506,107 @@ az aks create                      \
...
+                                            // Create EKS cluster: requires AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY environment
+                                            // variables to reference a user in AWS with the correct privileges to create an EKS
+                                            // cluster: https://github.com/weaveworks/eksctl/issues/204#issuecomment-450280786
+                                            container('eksctl') {
+                                                sh "eksctl create cluster --name ${clusterName} --version ${kversion} --node-type m5.large --nodes 3 --nodes-min 3 --nodes-max 3 --timeout 60m --kubeconfig \$KUBECONFIG --tags 'platform=${platform},branch=${BRANCH_NAME},build=${BUILD_URL}'"
+                                                withEnv(["PATH+KTOOL=${tool 'kubectl'}"]) {
+                                                    sh "kubectl create clusterrolebinding cluster-system-anonymous --clusterrole cluster-admin --user system:anonymous"
```
anguslees: Erm .. I think this line is granting `cluster-admin` to a huge number of cluster users, which at least deserves a long explanatory comment.

Why is this necessary?  Is there anything else we can do instead?

Jenkinsfile
anguslees: Also: we should run the tests on the same side of the planet as jenkins, or all requests to/from the cluster are unnecessarily slow :stuck_out_tongue: 
Best would be us-east-something.

Jenkinsfile
```diff
@@ -490,6 +506,107 @@ az aks create                      \
...
+                                                }
+                                                waitForRollout("kube-system", 30)
+                                            }
+
+                                            // NB: writeJSON doesn't work without approvals(?)
+                                            // See https://issues.jenkins-ci.org/browse/JENKINS-44587
```
anguslees: Not a relevant comment here - this isn't attempting to write a JSON file.

Jenkinsfile
```diff
@@ -490,6 +506,107 @@ az aks create                      \
...
+}
+"""
+                                            ])
+                                        }{
+                                            // update glue records in parent zone
+                                            def nameServers = []
+                                            container('aws') {
+                                                withEnv(["PATH+JQ=${tool 'jq'}"]) {
+                                                    def output = sh(returnStdout: true, script: """
+id=\$(aws route53 list-hosted-zones-by-name --dns-name "${dnsZone}" --max-items 1 | jq -r .HostedZones[0].Id)
```
anguslees: You can drop the `jq` requirement with:

```suggestion
id=\$(aws route53 list-hosted-zones-by-name --dns-name "${dnsZone}" --max-items 1 --query 'HostedZones[0].Id' --output text)
```

Jenkinsfile
```diff
@@ -490,6 +506,107 @@ az aks create                      \
...
+aws route53 get-hosted-zone --id "\$id" --query DelegationSet.NameServers
+"""
+                                                    )
+                                                    for (ns in readJSON(text: output)) {
```
anguslees: Why is this a loop?  I think we can just do `nameServers = readJSON(text: output)`

Jenkinsfile
```diff
@@ -122,6 +125,10 @@ def runIntegrationTest(String description, String kubeprodArgs, String ginkgoArg
                         junit 'junit/*.xml'
                     }
                 }
+
+                withEnv(["PATH+KUBECFG=${tool 'kubecfg'}"]) {
+                    sh "kubecfg -v delete kubeprod-manifest.jsonnet"
```
anguslees: Is this useful, or does it just introduce delays?  We're about to delete the entire cluster anyway, it would be nice if that was sufficient..

(Or: if we _need_ to delete all the contents of the cluster first, then presumably we need to wait for the implications of that to complete - wait until PVCs are unbound and PVs are removed, loadbalancers are deleted, etc.  Just doing this delete here and then immediately continuing seems inadequate.)

Jenkinsfile
```diff
@@ -490,6 +506,107 @@ az aks create                      \
...
+                                            sh "eksctl delete cluster --name ${clusterName} --timeout 60m"
+                                        }
+                                        container('aws') {
+                                            withEnv([
+                                                "PATH+JQ=${tool 'jq'}",
+                                                "HOME=${env.WORKSPACE}",
+                                            ]) {
+                                                sh """
+CONFIG="${clusterName}-autogen.json"
+ACCOUNT=\$(aws sts get-caller-identity | jq -r .Account)
```
anguslees:
```suggestion
ACCOUNT=\$(aws sts get-caller-identity --query Account --output text)
```

Jenkinsfile
```diff
@@ -490,6 +506,107 @@ az aks create                      \
...
+DNS_ZONE=\$(cat \${CONFIG} | jq -r .dnsZone)
```
anguslees: Is this just `${dnsZone}`?  If so, we may as well just use the groovy variable directly.

Jenkinsfile
falfaro: Deleting the cluster using `eksctl` doesn't seem enough: it does not remove any ELB interfaces which prevent the CloudFormation stacks from being deleted. If we are to wait for the cluster to be deleted, how can I do that?

Jenkinsfile
falfaro: Done.

Jenkinsfile
falfaro: Done.

Jenkinsfile
falfaro: It is necessary because, without this, the tests for reasons I don't know yet, run as an anonymous user. I tested that this is the case even from my laptop. Even when `kubectl` and `kubecfg` work fine. So there must be something in the initialization code in the tests that create an anonymous Kubernetes client.

Jenkinsfile
falfaro: Done.

Jenkinsfile
falfaro: Done.

Jenkinsfile
falfaro: Done.

Jenkinsfile
falfaro: Done.

Jenkinsfile
falfaro: Yes

Jenkinsfile
falfaro: Done

Jenkinsfile
falfaro: Updating the Kubernetes Go client fixed this.

Jenkinsfile
falfaro: Done.

Jenkinsfile
anguslees: Your `waitForDestroy()` looks like a reasonable solution.

Jenkinsfile
```diff
@@ -490,6 +532,115 @@ az aks create                      \
...
+                                        container('eksctl') {
+                                            sh "eksctl delete cluster --name ${clusterName} --timeout 60m"
+                                        }
+                                        container('aws') {
+                                            withEnv([
+                                                "PATH+JQ=${tool 'jq'}",
+                                                "HOME=${env.WORKSPACE}",
+                                            ]) {
+                                                sh """
+                                                    #!/bin/bash
```
anguslees: Add `set -xe`  (jenkins does this automatically with the default `sh` shell, but not if you provide your own #! line)

Jenkinsfile
```diff
@@ -490,6 +532,115 @@ az aks create                      \
...
+                                                    aws iam delete-policy --policy-arn "arn:aws:iam::\${ACCOUNT}:policy/bkpr-${dnsZone}"
+                                                    ACCESS_KEY_ID=\$(cat \${CONFIG} | jq -r .externalDns.aws_access_key_id)
+                                                    aws iam delete-access-key --user-name "bkpr-${dnsZone}" --access-key-id "\${ACCESS_KEY_ID}"
+                                                    aws iam delete-user --user-name "bkpr-${dnsZone}"
+                                                    CLIENT_ID=\$(cat \${CONFIG} | jq -r .oauthProxy.client_id)
+                                                    aws cognito-idp delete-user-pool-client --user-pool-id "${awsUserPoolId}" --client-id "\${CLIENT_ID}"
+                                                    DNS_ZONE_ID=\$(aws route53 list-hosted-zones-by-name --dns-name "${dnsZone}" --max-items 1 --query 'HostedZones[0].Id' --output text)
+                                                    aws route53 list-resource-record-sets --hosted-zone-id \${DNS_ZONE_ID} | jq -c '.ResourceRecordSets[]' | while read -r resourcerecordset; do
```
anguslees: I haven't tested it in anger, but I think the following is the same fwiw.  (apparently `--query` can do all sorts of things!)

```sh
aws route53 list-resource-record-sets \
  --hosted-zone-id \${DNS_ZONE_ID} \
  --query '{ChangeBatch:{Changes:ResourceRecordSets[?Type != `NS` && Type != `SOA`].{Action:`DELETE`,ResourceRecordSet:@}}}' \
  --output json > changes

aws route53 change-resource-record-sets \
  --cli-input-json file://changes \
  --hosted-zone-id \${DNS_ZONE_ID} \
  --query 'ChangeInfo.Id' \
  --output text

aws route53 delete-hosted-zone \
  --id \${DNS_ZONE_ID} \
  --query 'ChangeInfo.Id' \
  --output text
```

Jenkinsfile
falfaro: Done.

Jenkinsfile
falfaro: Works!

Jenkinsfile
falfaro: Done.

falfaro:
PTAL

Jenkinsfile
anguslees: Oops, sorry.  Looks like `readJSON` returns a `JSONArray`, but the `insertGlueRecords` function requires an `ArrayList`.  Need to put the loop back (boo!), or change `insertGlueRecords` to take a `java.util.List` (yay!) instead.

Jenkinsfile
sameersbn: i guess reverted :slightly_smiling_face:

Jenkinsfile
```diff
@@ -490,6 +499,150 @@ az aks create                      \
...
+                                        if (fileExists("${env.WORKSPACE}/.kubeconf")) {
+                                            // Uninstall BKPR
+                                            withEnv([
+                                                "PATH+KUBECFG=${tool 'kubecfg'}",
+                                                "PATH+KUBECTL=${tool 'kubectl'}",
+                                                "KUBECONFIG=${env.WORKSPACE}/.kubeconf",
+                                            ]) {
+                                                // This is required for AWS in order to release/destroy ELB network interfaces.
+                                                sh """
+                                                    #!/bin/bash
```
sameersbn: is this shebang really needed? and also the following `set` commands. I believe `set -x` is the default

Jenkinsfile
```diff
@@ -85,7 +86,7 @@ String gcpLabel(String s) {
 }
 
 def runIntegrationTest(String description, String kubeprodArgs, String ginkgoArgs, Closure clusterSetup, Closure dnsSetup) {
-    timeout(120) {
+    timeout(240) {
```
sameersbn: just noticed this, can it really take 4 hours for the tests to complete. Is this a realistic number?

Jenkinsfile
```diff
@@ -84,45 +85,48 @@ String gcpLabel(String s) {
...
+                } catch (error) {
+                    sh "kubectl --namespace kubeprod get po,deploy,svc,ing"
+                    input 'Paused for manual debugging'
+                    throw error
+                } finally {
+                    junit 'junit/*.xml'
                 }
             }
+
+            clusterDestroy()
```
anguslees: With this change, `clusterDestroy()` will not be run when any of the previous steps fail (raise an exception).  I think we need to run `clusterDestroy()` in some sort of try/finally block (see the previous code structure).

Jenkinsfile
anguslees: It is not needed here (afaics), and we should revert to standard `sh` (without the shebang, and without explicit `set +x`/`set -x` (?!))

Jenkinsfile
```diff
@@ -490,6 +502,152 @@ az aks create                      \
...
+                                                """
+
+                                                // Destroy AWS objects
+                                                container('aws') {
+                                                    withEnv([
+                                                        "PATH+JQ=${tool 'jq'}",
+                                                        "HOME=${env.WORKSPACE}",
+                                                    ]) {
+                                                        sh """
+                                                            #!/bin/bash
```
anguslees: Ditto, I don't believe this contains any bash-specific constructs anymore (an earlier version used `<<<`)

falfaro: Fixed.

Jenkinsfile
falfaro: Reverted to 120.

Jenkinsfile
falfaro: Wrapped inside a finally {}.

Jenkinsfile
falfaro: Fixed.

Jenkinsfile
```diff
@@ -255,7 +270,7 @@ spec:
         }, failFast: true)
 
     def platforms = [:]
-
+/*
```
anguslees: Still need to restore testing on other platforms.  You may need to change their cleanup code to follow your `try/finally` refactor.

falfaro: I'm on it :-)

Jenkinsfile
falfaro: Fixed.

falfaro:
@anguslees @sameersbn PTAL

tests/vendor/vendor.json
```diff
@@ -2,6 +2,10 @@
 	"comment": "",
 	"ignore": "test appengine",
 	"package": [
+		{
```
anguslees: This looks like the result of an incorrect govendor invocation.  We should remove this.

tests/vendor/vendor.json
```diff
@@ -44,6 +48,11 @@
 			"revision": "de5bf2ad457846296e2031421a34e2568e304e35",
 			"revisionTime": "2017-08-10T14:37:23Z"
 		},
+		{
```
anguslees: ditto. We should remove this.

falfaro: Fixed.

tests/vendor/vendor.json
falfaro: Fixed.

falfaro:
@anguslees @sameersbn please take another look as I think this is now robust enough to be merged.

sameersbn:
bors retry

sameersbn:
bors retry

Jenkinsfile
```diff
@@ -111,7 +111,9 @@ def runIntegrationTest(String description, String kubeprodArgs, String ginkgoArg
                     sh "kubectl --namespace kube-system get po,deploy,svc,ing"
 
                     try {
-                        sh "./bin/kubeprod -v=1 install ${kubeprodArgs} --manifests=manifests"
+                        retry(3) {
```
anguslees: Instead, is it sufficient to run `kubectl api-versions` somewhere above - and loop/delay until that succeeds?  Hopefully that will delay long enough that whatever `kubeprod install` is hitting will have stabilised.  Again, I would rather we didn't treat `kubeprod install` specially at all, because flakiness in this command is flakiness that the user will see (and we should instead fix `kubeprod install`).

sameersbn: hmm.. that makes sense. i'll make this change

PR #369, opened 2019-02-01
ci: prioritize testing on GKE over AKS

sameersbn:
It has been observed when there are a lot of parallel jobs in the
jenkins queue, jenkins does not appear to node assignments based on FIFO
basis. This results in the jobs taking a really long time to complete
(courtesy the AKS start up time).

Since the GKE clusters come up quickly, so do the tests complete faster.
This PR is a workaround to prioritize testing on GKE clusters so that
problematic PR's fails faster in such cases

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -254,6 +254,112 @@ spec:
     def platforms = [:]
 
     // See:
+    //  gcloud container get-server-config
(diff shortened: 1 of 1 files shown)


falfaro:
LGTM

PR #403, opened 2019-02-21
Allow oauth2_proxy to be used from other namespaces

### LINKED ISSUE #317: Reusing BKPR oauth2 setup for custom applications
I want to use KBPR's components like the oauth2-proxy for one of my applications. By configuring an ingress resource I want to route a call to myapp.mydomain.com/oauth2 to the oauth2-proxy backend in the kubepord namespace.

The problem is my application is running in the `default` namespace. Not in the `kubeprod` one. As a result, my ingress resource needs to be defined in the `default` namespace as well. As a consequence, I cannot link calls to `/oauth2` to the oauth2-proxy service b/c that's running in the `kubeprod` namespace.

Is it one of the objectives of KBPR to reuse components for custom applications? 
If so, should I just put my application in the `kubeprod` namespace? 
Should I use another oauth2-proxy in the `default` namespace instead?
Is there another way?

anguslees:
Improve the oauth2_proxy setup to be able to protect Ingress objects
from other k8s namespaces too.

This change will protect any Ingress, provided it:
a) uses a hostname within the zone managed by BKPR, and
b) includes the required Ingress annotations (see below).

The key changes involve:
- oauth2_proxy (encrypted) cookie is now exposed to the full DNS
  zone given by `--dns-zone`
- Upstream identity provider setup is simplified to only redirect to
  https://auth.${dns_zone}/oauth2/callback
- oauth2_proxy setup is modified to perform redirects to any host
  within ${dns_zone}

To protect your Ingress with the BKPR-managed oauth2 stack, your
Ingress requires the following annotations (with `${dns_zone}`
replaced accordingly. Note other $... strings should remain as-is).
If using jsonnet, deriving from `utils.AuthIngress(name)` will do the
right thing.
```
annotations:
  kubernetes.io/ingress.class: nginx
  nginx.ingress.kubernetes.io/auth-signin: https://auth.${dns_zone}/oauth2/start?rd=%2F$server_name$escaped_request_uri
  nginx.ingress.kubernetes.io/auth-url: https://auth.${dns_zone}/oauth2/auth
  # optional, if you want access to the authenticated user/email name
  nginx.ingress.kubernetes.io/auth-response-headers: X-Auth-Request-User, X-Auth-Request-Email
```

### Upgrade impact:
The Oauth2 redirect URI configured in cloud setups needs to be changed
to `https://auth.${dns_zone}/oauth2/callback` (with `${dns_zone}` replaced
appropriately).  Other redirect URIs will no longer be used after this
change, and may be removed (they are also safe to leave configured).

Fixes #317

### DIFF
--- a/docs/quickstart-gke.md
+++ b/docs/quickstart-gke.md
@@ -84,6 +84,7 @@ In this section, you will deploy a Google Kubernetes Engine (GKE) cluster using 
       + https://prometheus.${BKPR_DNS_ZONE}/oauth2/callback
       + https://kibana.${BKPR_DNS_ZONE}/oauth2/callback
       + https://grafana.${BKPR_DNS_ZONE}/oauth2/callback
+      + https://auth.${BKPR_DNS_ZONE}/oauth2/callback
 
   > Replace `${BKPR_DNS_ZONE}` with the value of the `BKPR_DNS_ZONE` environment variable*
 
--- a/manifests/components/kibana.jsonnet
+++ b/manifests/components/kibana.jsonnet
@@ -66,11 +66,7 @@ local strip_trailing_slash(s) = (
               },
               env_+: {
                 KIBANA_ELASTICSEARCH_URL: $.es.svc.host,
-
-                local route = $.ingress.spec.rules[0].http.paths[0],
-                // Make sure we got the correct route
-                assert route.backend == $.svc.name_port,
-                SERVER_BASEPATH: strip_trailing_slash(route.path),
+                SERVER_BASEPATH: strip_trailing_slash($.ingress.kibanaPath),
                 KIBANA_HOST: "0.0.0.0",
                 XPACK_MONITORING_ENABLED: "false",
                 XPACK_SECURITY_ENABLED: "false",
@@ -92,13 +88,14 @@ local strip_trailing_slash(s) = (
   ingress: utils.AuthIngress($.p + "kibana-logging") + $.metadata {
     local this = self,
     host:: error "host is required",
+    kibanaPath:: "/",
     spec+: {
       rules+: [
         {
           host: this.host,
           http: {
             paths: [
-              { path: "/", backend: $.svc.name_port },
+              { path: this.kibanaPath, backend: $.svc.name_port },
             ],
           },
         },
--- a/manifests/components/kibana.jsonnet
+++ b/manifests/components/kibana.jsonnet
@@ -67,7 +67,7 @@ local strip_trailing_slash(s) = (
               env_+: {
                 KIBANA_ELASTICSEARCH_URL: $.es.svc.host,
 
-                local route = $.ingress.spec.rules[1].http.paths[0],
+                local route = $.ingress.spec.rules[0].http.paths[0],
                 // Make sure we got the correct route
                 assert route.backend == $.svc.name_port,
                 SERVER_BASEPATH: strip_trailing_slash(route.path),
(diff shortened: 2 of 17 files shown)


docs/quickstart-gke.md
```diff
@@ -84,6 +84,7 @@ In this section, you will deploy a Google Kubernetes Engine (GKE) cluster using
       + https://prometheus.${BKPR_DNS_ZONE}/oauth2/callback
       + https://kibana.${BKPR_DNS_ZONE}/oauth2/callback
       + https://grafana.${BKPR_DNS_ZONE}/oauth2/callback
+      + https://auth.${BKPR_DNS_ZONE}/oauth2/callback
```
falfaro: Why don't you remove the three previous lines, as they won't be needed anymore IIUC?

manifests/components/kibana.jsonnet
```diff
@@ -67,7 +67,7 @@ local strip_trailing_slash(s) = (
               env_+: {
                 KIBANA_ELASTICSEARCH_URL: $.es.svc.host,
 
-                local route = $.ingress.spec.rules[1].http.paths[0],
+                local route = $.ingress.spec.rules[0].http.paths[0],
```
falfaro: Isn't there a way to make this more robust than using an index? This is not immediately obvious to the casual reader. So at least it deserves a nice comment.

manifests/platforms/aks.jsonnet
```diff
@@ -99,6 +99,10 @@ local grafana = import "../components/grafana.jsonnet";
       data_+: $.config.oauthProxy,
     },
 
+    ingress+: {
```
falfaro: Can't we have this in a generic jsonnet file instead of three times (for each platform-specific jsonnet)?

manifests/platforms/eks.jsonnet
```diff
@@ -98,6 +98,10 @@ local grafana = import "../components/grafana.jsonnet";
       data_+: $.config.oauthProxy,
     },
 
+    ingress+: {
```
falfaro: Ditto.

manifests/platforms/gke.jsonnet
```diff
@@ -99,6 +99,10 @@ local grafana = import "../components/grafana.jsonnet";
       data_+: $.config.oauthProxy,
     },
 
+    ingress+: {
```
falfaro: Ditto.

docs/quickstart-gke.md
anguslees: It was too hard to explain that the old URLs are only needed for releases before this PR, and that in the future (after this PR) they will need to go back and change the URLs (we can't do this automatically because Google has no API for oauth2 setup).  I couldn't come up with a solution that was any better than "just list all of them" - if we had a system of versioned docs then there might have been other options available.

Note: We can remove the old URLs after we have made a release containing this PR, but again the cost of leaving them is very low.

manifests/components/kibana.jsonnet
anguslees: replaced with something more explicit

manifests/platforms/aks.jsonnet
anguslees: Agreed, we should refactor the common parts out of all platforms into a platforms/common.libsonnet or something.  I think originally I expected greater deviation between platforms.

I would rather do that refactor in a separate PR, after this PR.  I can do that before (or within this PR) if you would prefer.

PR #405, opened 2019-02-22
GKE no longer supports 1.10

anguslees:
https://cloud.google.com/kubernetes-engine/release-notes#february-18-2019

```
Versions no longer available
The following versions are no longer available for new clusters or
upgrades to existing cluster masters:

1.10.x
```

This means GKE currently _only_ supports 1.11.x (masters) :/

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -258,7 +258,7 @@ spec:
 
     // See:
     //  gcloud container get-server-config
-    def gkeKversions = ["1.10", "1.11"]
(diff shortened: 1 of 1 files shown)


sameersbn:
While this fixes the issue for now, I was thinking whether it would be better if we simply queried each available k8s version on each cloud and test the builds on all available versions. This would mean that we could be testing bkpr against k8s versions that are not advertised to be supported.

PR #406, opened 2019-02-22
external-dns: Add trailing '.' to txt-prefix

anguslees:
This is an improvement/fix to #343, to achieve the originally stated
goal.  external-dns prepends _exactly_ the txt-prefix to created TXT
metadata records, so the `.` separator needs to be added explicitly.


anguslees:
bors r=sameersbn

PR #408, opened 2019-02-25
Update roadmap with the plans for 1.2

arapulido:


docs/roadmap.md
```diff
@@ -4,8 +4,9 @@ Caveat: contains forward-looking statements that may or may not match reality. [
 
 ## Release 1.2
 
-- Support for AWS/kops platform ([#292](https://github.com/bitnami/kube-prod-runtime/issues/292))
```
anguslees: We should bump this to 1.3, rather than drop it entirely. (imo - assuming that's our plan)

arapulido: Good point. Added back to 1.3

PR #413, opened 2019-03-01
Update deployment when ext-dns or oauth2_proxy secret changes

### LINKED ISSUE #64: external-dns doesn't reread updated token secret
Needs an explicit restart (kill the pod) to pick up the new (AKS) secret.  We need to add a config-reloader sidecar, hash the secret name, fix the code, etc.

anguslees:
Add new `HashedConfigMap` and `HashedSecret` derived objects that
modify their `metadata.name` based on a hash of `self.data`.

Change external-dns and oauth2_proxy `Secret`s to `HashedSecret`,
which will force a `Deployment` update whenever the contents (name) of
the referenced `Secret` changes.

Going forward, we should use this pattern for any processes that don't
re-read a modified mounted file and use a configmap/secret that is
static after creation.

Fixes #64

### DIFF
--- a/manifests/components/oauth2-proxy.jsonnet
+++ b/manifests/components/oauth2-proxy.jsonnet
@@ -31,7 +31,7 @@ local OAUTH2_PROXY_IMAGE = (import "images.json")["oauth2_proxy"];
     },
   },
 
-  secret: kube.Secret($.p + "oauth2-proxy") + $.metadata {
+  secret: utils.HashedSecret($.p + "oauth2-proxy") + $.metadata {
     data_+: {
       client_id: error "client_id is required",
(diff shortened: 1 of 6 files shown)


arapulido:
@jjo

anguslees:
Aha, I had a feeling I was relying on the configmap name somewhere.  Updated the integration test to follow the new rename.

anguslees:
I removed the `backports/queue` tag just now after discussion.  This change is a bugfix, and small in terms of jsonnet change, **but** it changes the names of several important k8s resources (to add the hash suffix).  This last seems a scary thing to slide into an existing release (and against semver, etc).

PR #414, opened 2019-03-01
Increase nginx-ingress + oauth2_proxy to replicas=2

### LINKED ISSUE #373: nginx-ingress: Increase availability during node failure
Is the quick start intended to be HA?

I think it'd make sense to have a HA configuration - e.g. the default GKE quickstart only has one replica of nginx-ingress which will involve down-time during node upgrades. Or at least have a separate `kubeprod-manifest.jsonnet` with the over-rides.

anguslees:
nginx-ingress and oauth2_proxy are expected to be in the serving path,
and so should be redundant to single-machine failure (as per the BKPR
default HA target).

Accordingly:
- increase replicas for each to 2
- add (weak) host anti-affinity
- configure a `PodDisruptionBudget` that ensures at least 1 replica of
  each is healthy before draining a node.

Fixes #373

### DIFF
--- a/manifests/components/nginx-ingress.jsonnet
+++ b/manifests/components/nginx-ingress.jsonnet
@@ -18,6 +18,7 @@
  */
 
 local kube = import "../lib/kube.libsonnet";
+local utils = import "../lib/utils.libsonnet";
 
 local NGNIX_INGRESS_IMAGE = (import "images.json")["nginx-ingress-controller"];
 
@@ -141,11 +142,23 @@ local NGNIX_INGRESS_IMAGE = (import "images.json")["nginx-ingress-controller"];
 
   hpa: kube.HorizontalPodAutoscaler($.p + "nginx-ingress-controller") + $.metadata {
     target: $.controller,
-    spec+: {maxReplicas: 10},
+    spec+: {
+      // Put a cap on growth due to (eg) DoS attacks.
+      // Large sites will want to increase this to cover legitimate demand.
+      maxReplicas: 10,
+    },
+  },
+
+  pdb: kube.PodDisruptionBudget($.p + "nginx-ingress-controller") + $.metadata {
+    target_pod: $.controller.spec.template,
+    spec+: {minAvailable: $.controller.spec.replicas - 1},
   },
 
   controller: kube.Deployment($.p + "nginx-ingress-controller") + $.metadata {
+    local this = self,
     spec+: {
+      // Ensure at least n+1.  NB: HPA will increase replicas dynamically.
+      replicas: 2,
       template+: {
         metadata+: {
           annotations+: {
@@ -158,6 +171,7 @@ local NGNIX_INGRESS_IMAGE = (import "images.json")["nginx-ingress-controller"];
           serviceAccountName: $.serviceAccount.metadata.name,
           //hostNetwork: true, // access real source IPs, IPv6, etc
           terminationGracePeriodSeconds: 60,
+          affinity+: utils.weakNodeDiversity(this.spec.selector),
           containers_+: {
             default: kube.Container("nginx") {
               image: NGNIX_INGRESS_IMAGE,
--- a/manifests/components/oauth2-proxy.jsonnet
+++ b/manifests/components/oauth2-proxy.jsonnet
@@ -18,6 +18,7 @@
  */
 
 local kube = import "../lib/kube.libsonnet";
+local utils = import "../lib/utils.libsonnet";
 
 local OAUTH2_PROXY_IMAGE = (import "images.json")["oauth2_proxy"];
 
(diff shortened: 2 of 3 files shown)


manifests/components/nginx-ingress.jsonnet
```diff
@@ -141,11 +141,23 @@ local NGNIX_INGRESS_IMAGE = (import "images.json")["nginx-ingress-controller"];
...
+    spec+: {
+      // Put a cap on growth due to (eg) DoS attacks.
+      // Large sites will want to increase this to cover legitimate demand.
+      maxReplicas: 10,
+    },
+  },
+
+  pdb: kube.PodDisruptionBudget($.p + "nginx-ingress-controller") + $.metadata {
+    target_pod: $.controller.spec.template,
+    spec+: {minAvailable: $.controller.spec.replicas - 1},
```
falfaro: This only allows one instance to be down at a time, right? Does NGINX support lame-duck mode?

manifests/components/nginx-ingress.jsonnet
```diff
@@ -141,11 +141,23 @@ local NGNIX_INGRESS_IMAGE = (import "images.json")["nginx-ingress-controller"];
 
   hpa: kube.HorizontalPodAutoscaler($.p + "nginx-ingress-controller") + $.metadata {
     target: $.controller,
-    spec+: {maxReplicas: 10},
+    spec+: {
+      // Put a cap on growth due to (eg) DoS attacks.
+      // Large sites will want to increase this to cover legitimate demand.
+      maxReplicas: 10,
```
falfaro: Do we need a minReplicas: 2 or some such to prevent auto-scaling from messing things up?

manifests/components/oauth2-proxy.jsonnet
```diff
@@ -44,13 +44,36 @@ local OAUTH2_PROXY_IMAGE = (import "images.json")["oauth2_proxy"];
 
   hpa: kube.HorizontalPodAutoscaler($.p + "oauth2-proxy") + $.metadata {
     target: $.deploy,
-    spec+: {maxReplicas: 10},
+    spec+: {
+      // Put a cap on growth due to (eg) DoS attacks.
+      // Large sites will want to increase this to cover legitimate demand.
+      maxReplicas: 10,
```
falfaro: Do we need a minReplicas: 2 or some such to prevent auto-scaling from messing things up?

manifests/components/nginx-ingress.jsonnet
```diff
@@ -158,6 +170,17 @@ local NGNIX_INGRESS_IMAGE = (import "images.json")["nginx-ingress-controller"];
           serviceAccountName: $.serviceAccount.metadata.name,
           //hostNetwork: true, // access real source IPs, IPv6, etc
           terminationGracePeriodSeconds: 60,
+          affinity+: {
```
falfaro: A comment that describes what this does would be very helpful.

manifests/components/oauth2-proxy.jsonnet
```diff
@@ -44,13 +44,36 @@ local OAUTH2_PROXY_IMAGE = (import "images.json")["oauth2_proxy"];
...
   },
 
   deploy: kube.Deployment($.p + "oauth2-proxy") + $.metadata {
+    local this = self,
     spec+: {
+      // Ensure at least n+1.  NB: HPA will increase replicas dynamically.
+      replicas: 2,
       template+: {
         spec+: {
+          affinity+: {
```
falfaro: A comment that describes what this does would be very helpful.

manifests/components/nginx-ingress.jsonnet
anguslees: nginx-ingress controller supports lame-duck, and we configure the nginx pod with `terminationGracePeriodSeconds=60`.
Since I just checked, the controller is running as pid 1, so it will receive the SIGTERM, and it [catches that](https://github.com/kubernetes/ingress-nginx/blob/17e788b8e11536e037e52cfd56e5849e7bb5405f/cmd/nginx/main.go#L163) and turns it into a clean nginx shutdown.  It's not immediately clear to me [from the code](https://github.com/kubernetes/ingress-nginx/blob/17e788b8e11536e037e52cfd56e5849e7bb5405f/internal/ingress/controller/checker.go#L38) whether it fails healthz checks during that time or not.  (Actual observations welcome)

manifests/components/nginx-ingress.jsonnet
anguslees: `kube.libsonnet` already inserts a `minReplicas=$.target.spec.replicas` for us.

manifests/components/oauth2-proxy.jsonnet
anguslees: ditto

manifests/components/nginx-ingress.jsonnet
anguslees: Since it's something we're likely to use a few times, I moved it into a more self-descriptive `utils.weakNodeDiversity`.

For anyone reading the PR archives, the relevant upstream anti-affinity docs are https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity

manifests/components/oauth2-proxy.jsonnet
anguslees: ditto.

PR #415, opened 2019-03-01
[jjo] add contrib/helpers.jsonnet, including unit tests

jjo:
See `contrib/helpers_usage_test.jsonnet` for example usage.

### DIFF
--- a/manifests/contrib/showtree.jsonnet
+++ b/manifests/contrib/showtree.jsonnet
@@ -29,7 +29,7 @@ local formatKube(k, obj) = (
 // Build main tree: jsonnet key -> formatted Kubernetes object
 local show_tree(obj) = (
   local tree(key, x) = (
-    if std.objectHas(x, 'metadata') then
+    if std.objectHas(x, 'apiVersion') && std.objectHas(x, 'kind') then
       formatKube(key, x)
     else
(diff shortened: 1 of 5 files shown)


arapulido:
Maybe would it be good to add some documentation of those to: https://github.com/bitnami/kube-prod-runtime/blob/master/docs/overrides.md

manifests/contrib/showtree.jsonnet
```diff
@@ -0,0 +1,54 @@
...
+  if std.objectHas(obj.metadata, 'namespace') then
+    { [k]: '%s %s -n %s' % [obj.kind, obj.metadata.name, obj.metadata.namespace] }
+  else
+    { [k]: '%s %s' % [obj.kind, obj.metadata.name] }
+);
+
+// Build main tree: jsonnet key -> formatted Kubernetes object
+local show_tree(obj) = (
+  local tree(key, x) = (
+    if std.objectHas(x, 'metadata') then
```
anguslees:
```suggestion
    if std.objectHas(x, 'apiVersion') && std.objectHas(x, 'kind') then
```

(checking both is probably unnecessarily strict)

anguslees:
Merging.  We can think about if/how we promote/expose this later.

bors r+

PR #430, opened 2019-03-12
Workaround for crashing Elasticsearch 6.x under EKS.

falfaro:


docs/quickstart-eks.md
```diff
@@ -60,6 +60,15 @@ In this section, you will deploy an Amazon Elastic Container Service for Kuberne
                         --nodes=3 \
                         --version=${AWS_EKS_K8S_VERSION}
   ```
+  > **NOTE**: At the time of this writing, EKS clusters created with `eksctl` are affected by a [bug](https://github.com/awslabs/amazon-eks-ami/issues/193) that causes Elasticsearch to get into a crashloop. A temporary workaround consists of overriding the AMI used when creating the cluster. The AMI named `amazon-eks-node-1.10-v20190211` is known to work. You will need to find its ID that corresponds to the region and zone where you are creating the cluster. For instance, in `eu-central-1`, the AMI ID is `ami-074583f8d5a05e27b`. Use the `--node-ami` command-line argument to `eksctl` to override the AMI. For example:
+
```
wojciechka: I'd at least also list us-east-1 AMI as this is where most people deploy to.

falfaro: Done.

sameersbn:
IMO we should instead ask users to SSH into their EKS cluster nodes and remove the ulimits configuration of the docker daemon. This can be achieved with:

```bash
sudo sed -i '/"nofile": {/,/}/d' /etc/docker/daemon.json
sudo systemctl restart docker
```

This workaround can go in the troubleshooting doc instead of the quickstart guide.

falfaro:
@sameersbn in my opinion, we should make things as easy as possible for users, and adding `--node-ami` to tke `eksctl` command-line is way easier than SSH-ing into N machines, changing the Docker configuration, etc.

Also, if the change is going to be reverted soon, what is the point of adding this documentation to the troubleshooting section? Instead, we can keep it in the Quickstart and once Amazon fixes the issue, we can remove this block of documentation.

@arapulido what do you think?

sameersbn:
This workaround only works for newly created clusters. User with an existing EKS cluster will not be able to take the path of this workaround.

wojciechka:
> This workaround only works for newly created clusters. User with an existing EKS cluster will not be able to take the path of this workaround.

That's true. What I managed to do is create a new nodegroup with proper AMI and then `kubectl drain` all my old nodes, then deleting nodegroup.

I think `eksctl` even helps with that - i.e. https://github.com/weaveworks/eksctl/pull/592

It was a painful process and mongodb that I had installed for my kubeapps could not be moved for some reason - but I just uninstalled and reinstalled kubeapps.

arapulido:
I think we should have both. Pointing to a previous AMI in the quickstart (and revert this once it is fixed), and a troubleshooting section to manually change the ulimits that can stay (for people who already have a EKS cluster)

docs/quickstart-eks.md
```diff
@@ -60,6 +60,22 @@ In this section, you will deploy an Amazon Elastic Container Service for Kuberne
                         --nodes=3 \
                         --version=${AWS_EKS_K8S_VERSION}
   ```
+  > **NOTE**: At the time of this writing, EKS clusters created with `eksctl` are affected by a [bug](https://github.com/awslabs/amazon-eks-ami/issues/193) that causes Elasticsearch to get into a crashloop. A temporary workaround consists of overriding the AMI used when creating the cluster. The AMI named `amazon-eks-node-1.10-v20190211` is known to work. You will need to find its ID that corresponds to the region and zone where you are creating the cluster. For instance:
```
anguslees: > You will need to find its ID

Can we include the relevant command that dumps out the below IDs table?

falfaro: Done.

falfaro: Done.

docs/troubleshooting.md
```diff
@@ -449,3 +451,53 @@ Non-authoritative answer:
 Name:   bitnami.com
 Address: 50.17.235.25
 ```
+
+## Troubleshooting Elasticsearch
+
+### Elasticsearch crashloop under EKS
```
sameersbn: Following the pattern used in the rest of the document where we first state the issue that's observed (i.e. Elasticsearch enters a crashloop) and then include a Troubleshooting section which would walk the user to confirm that the issue is due to the ulimits and then suggest using the `sed` command to resolve the issue.

falfaro: What about now?

docs/troubleshooting.md
```diff
@@ -456,7 +456,23 @@ Address: 50.17.235.25
...
-EKS clusters which use the AMI image named `amazon-eks-node-1.10-v20190220` in any nodegroup are known to be affected by a [bug](https://github.com/awslabs/amazon-eks-ami/issues/193) that causes Elasticsearch 6.0 to get into a crashloop. In order to fix this, you will SSH into every node from the EKS nodegroup and remove the `ulimit` configuration of the Docker daemon:
+When Elasticsearch enters into a never-ending crashloop under EKS, the logs can reveal the cause:
+
+```console
+$ kubectl --namespace=kubeprod logs elasticsearch-logging-0 elasticsearch-logging
+```
+
+__Troubleshooting__:
+
+If you see logged mesasges like this:
```
sameersbn: s/mesasges/messages/

docs/troubleshooting.md
```diff
@@ -456,7 +456,23 @@ Address: 50.17.235.25
...
+
+```
+[1]: max file descriptors [8192] for elasticsearch process is too low, increase to at least [65536]
+```
+
+Then, you are affected by a [bug](https://github.com/awslabs/amazon-eks-ami/issues/193) that causes Elasticsearch 6.0 to get into a crashloop because of incorrect `ulimit` values. It affects EKS clusters which use the AMI image named `amazon-eks-node-1.10-v20190220`.
+
+In order to fix this, you will SSH into every node from the EKS nodegroup and remove the `ulimit` configuration of the Docker daemon:
```
sameersbn: s/you will SSH into every node/you will need to SSH into every node/

sameersbn:
lgtm! minor typos to be resolved

docs/troubleshooting.md
falfaro: Fixed.

docs/troubleshooting.md
falfaro: Fixed.

PR #457, opened 2019-03-22
ci: retry failed stages only once after a failure

sameersbn:
Retrying failed build stages 3 times is excessive and wasteful in most
cases. To account for transient cluster creation failures, one
additional retry should be sufficient to overcome issues coming from
cluster creation issues.

Additionally we have added a 3 minute wait before retrying.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -277,7 +277,7 @@ spec:
                     }, failFast: true
                 )
 
-                def maxRetries = 3
+                def maxRetries = 2
                 def platforms = [:]
 
                 // See:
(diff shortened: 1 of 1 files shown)


sameersbn:
I have also noticed that the time it tasks for 3 retries is well over 120 mins (the job timeout). When this timeout is reached the job is aborted leaving the cluster resources behind. Instead of reducing the number of retries I could have increased the timeout to 3 hours. But i feel that if the job does not PASS in two attempts, there is a very very low chance that it'll PASS in the third.

sameersbn:
bors retry

PR #488, opened 2019-04-15
ci: remove use of pre-emptible vms in GKE

sameersbn:
We've been seeing frequent failure in the cluster set up on GKE in the
CI which are not observed during development. This PR tests if removing
the usage of pre-emptible vms makes any difference


sameersbn:
use of pre-emptible vm's does not appear to make any difference

PR #500, opened 2019-04-22
feat: Add a default set of Grafana dashboards 

dbarranco:
Hi!

This PR adds a set of 3 dashboards via ConfigMap for Grafana.
As it is the first set of default dashboards that we are adding, they aim to provide:

* Kubernetes cluster capacity:

Screenshot 2019-04-22 at 15 03 01

* Kubernetes resource usage per namespace, pod and container:

Screenshot 2019-04-22 at 15 03 24

* Kubernetes workload summary:

Screenshot 2019-04-22 at 15 04 06

----

As they are loaded via Configmaps, these dashboards will be automatically added under the `Kubernetes` folder:

Screenshot 2019-04-22 at 15 04 55

These dashboards are imported as a submodule of https://github.com/bitnami-labs/kubernetes-grafana-dashboards, where we will be maintaining them.
In the near future we will contribute with more dashboards related with other BKPR components, like the Elasticsearch cluster.

---
These dashboards have been tested on a fresh installation of BKPR in a GKE 1.11 cluster, Node_exporter 0.17.0.
Cheers.

PR #501, opened 2019-04-23
feat: Add customize guide for Grafana provisioned dashboards and add code improvements to components/grafana.jsonnet

dbarranco:
I was missing the guide to customise the Grafana provisioned dashboards, so it is added here.
Also, feedback that @anguslees gave me in #500 has been applied here.

I'm still thinking on a good way to test the Grafana dashboards, will come with another PR with more improvements :) 

Cheers

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -377,6 +377,60 @@ The following example shows how to override the amount of persistent storage req
 }
 ```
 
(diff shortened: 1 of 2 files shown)


sameersbn:
bors retry

PR #510, opened 2019-04-25
manifests: updated RBAC permissions for external-dns

sameersbn:
ref: https://github.com/kubernetes-incubator/external-dns/issues/990


falfaro:
LG, BTW!

PR #513, opened 2019-04-26
tests: [fix] retry monitoring/exporters test

sameersbn:


sameersbn:
first attempt of fixing this was a fail: https://github.com/bitnami/kube-prod-runtime/pull/502

tests/monitoring_test.go
```diff
@@ -92,40 +92,29 @@ var _ = Describe("Exporters", func() {
 	})
 
 	DescribeTable("Exported timeseries", func(selector string, match types.GomegaMatcher) {
+		var series []map[string]string
```
anguslees: I think this should be inside `Eventually(func() { ... })` too..

tests/monitoring_test.go
```diff
@@ -92,40 +92,29 @@ var _ = Describe("Exporters", func() {
...
+				Services("kubeprod").
+				ProxyGet("http", "prometheus", "9090", "api/v1/series", params).
+				DoRaw()
+			Expect(err).NotTo(HaveOccurred())
+
+			resp := promResponse{}
+			json.Unmarshal(resultRaw, &resp)
+			json.Unmarshal(resp.Data, &series)
+
+			fmt.Printf("%v\n", series)
```
anguslees: Don't write to stdout, write to `GinkgoWriter`.  That way the logging output is correctly associated with the test, and appears only if `ginkgo -v` is used.

tests/monitoring_test.go
```diff
@@ -92,40 +92,29 @@ var _ = Describe("Exporters", func() {
...
+
+			return series, err
+		}, "5m", "5s").Should(match)
 	},
-		Entry("prometheus", `prometheus_tsdb_head_chunks{kubernetes_namespace="kubeprod",name="prometheus"}`, Not(BeEmpty())),
+		Entry("prometheus", `prometheus_tsdb_head_chunks{kubernetes_namespace="kubeprod",name="none"}`, Not(BeEmpty())),
```
anguslees: Is this necessary? `name="none"` seems like it might be a bug somewhere.

tests/monitoring_test.go
sameersbn: done!

tests/monitoring_test.go
sameersbn: oops! this was not supposed to be committed

tests/monitoring_test.go
sameersbn: damn! should not have been there.

sameersbn:
@anguslees PTAL!

Issue #453, opened 2019-03-21
kibana: make it easier to install plugins


anguslees:
There is currently no _easy_ declarative method to install kibana plugins afaics :(

We should add an initContainer that runs `kibana-plugin install ...`, and add a shared plugins volume between that initContainer and the regular kibana container.   See the [helm chart](https://github.com/helm/charts/blob/8ca2d5d8c9cc8a1c828b23b4e57e9d32097635b4/stable/kibana/templates/deployment.yaml#L71) for inspiration.

_Originally posted by @anguslees in https://github.com/bitnami/kube-prod-runtime/issues/386#issuecomment-475081671_

sameersbn:
Implemented in #530

PR #530, opened 2019-05-06
Support for installing Kibana plug-ins

falfaro:
This pull requests adds support for a very rudimentary mechanism to install additional Kibana plug-ins via an init container.

https://github.com/bitnami/kube-prod-runtime/issues/453

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -150,6 +150,32 @@ Kibana connects to the Elasticsearch cluster via the `elasticsearch-logging` Kub
 
 Kibana is a stateless component and therefore does not have any persistent storage requirements.
 
+#### Customization
+
+##### Kibana plug-ins
+
+Add-on functionality for Kibana is implemented with plug-in modules. Some known plug-ins for Kibana are listed here: https://www.elastic.co/guide/en/kibana/6.7/known-plugins.html. Please ensure the plugins you install are compatible with the version of Kibana installed by your BKPR version.
+
+To install additional plug-ins, override the `plugins` item inside the `kibana` scope, like this:
+
+```jsonnet
+(import "../../manifests/platforms/aks.jsonnet") {
+    config:: import "kubeprod-autogen.json",
+    kibana+: {
+        plugins+: {
+            "enhanced-table": {
+                version: "1.2.0",
+                url: "https://github.com/fbaligand/kibana-enhanced-table/releases/download/v1.2.0/enhanced-table-1.2.0_6.7.2.zip",
+            },
+            "network_vis": {
+                version: "6.7.2-3",
+                url: "https://github.com/uniberg/kbn_sankey_vis/releases/download/6.7.2-3/sankey_vis-6.7.2-3.zip",
+            },
+        },
+    },
+}
+```
+
 ## Monitoring stack
 ### Prometheus
 
--- a/manifests/components/kibana.jsonnet
+++ b/manifests/components/kibana.jsonnet
@@ -22,6 +22,7 @@ local kubecfg = import "kubecfg.libsonnet";
 local utils = import "../lib/utils.libsonnet";
 
 local KIBANA_IMAGE = (import "images.json")["kibana"];
+local KIBANA_PLUGINS_PATH = "/opt/bitnami/kibana/plugins";
 
 local strip_trailing_slash(s) = (
   if std.endsWith(s, "/") then
@@ -38,9 +39,23 @@ local strip_trailing_slash(s) = (
     },
   },
 
+  // List of Kibana plug-ins to install
+  plugins:: {
+    /* Example:
+    logtrail: {
+      version: "0.1.31",
+      url: "https://github.com/sivasamyk/logtrail/releases/download/v0.1.31/logtrail-6.7.1-0.1.31.zip",
+    },
+    */
+  },
+
   es: error "elasticsearch is required",
 
   serviceAccount: kube.ServiceAccount($.p + "kibana") + $.metadata {
+  },
+
+  pvc: kube.PersistentVolumeClaim($.p + "kibana-plugins") + $.metadata {
+    storage: "2Gi",
   },
 
   deploy: kube.Deployment($.p + "kibana") + $.metadata {
@@ -49,6 +64,40 @@ local strip_trailing_slash(s) = (
         spec+: {
           securityContext: {
             fsGroup: 1001,
+          },
+          volumes_+: {
+            plugins: kube.PersistentVolumeClaimVolume($.pvc),
+          },
+          initContainers_+: {
+            kibana_plugins_install: kube.Container("kibana-plugins-install") {
+              image: KIBANA_IMAGE,
+              securityContext: {
+                allowPrivilegeEscalation: false,
+              },
+              local wanted = std.join("\n", ["%s@%s,%s" % [k, $.plugins[k].version, $.plugins[k].url] for k in std.objectFields($.plugins)]),
+              command: [
+                "/bin/sh",
+                "-c",
+                |||
+                  set -e
+                  rm -rf /opt/bitnami/kibana/plugins/lost+found
+                  echo %s | sort > /tmp/wanted.list
+                  /opt/bitnami/kibana/bin/kibana-plugin list | grep @ | sort > /tmp/installed.list
+                  join -v2 -t, -j1 /tmp/wanted.list /tmp/installed.list | while read plugin; do
+                    ${plugin:+/opt/bitnami/kibana/bin/kibana-plugin remove "${plugin%%@*}"}
+                  done
+                  join -v1 -t, -j1 -o1.2 /tmp/wanted.list /tmp/installed.list | while read url; do
+                    ${url:+/opt/bitnami/kibana/bin/kibana-plugin install --no-optimize "$url"}
+                  done
+                ||| % std.escapeStringBash(wanted)
+              ],
+              volumeMounts_+: {
+                // Persistence for Kibana plugins
+                plugins: {
+                  mountPath: KIBANA_PLUGINS_PATH,
+                },
+              },
+            },
           },
           containers_+: {
             kibana: kube.Container("kibana") {
@@ -73,6 +122,12 @@ local strip_trailing_slash(s) = (
               },
               ports_+: {
                 ui: { containerPort: 5601 },
+              },
+              volumeMounts_+: {
+                // Persistence for Kibana plugins
+                plugins: {
+                  mountPath: KIBANA_PLUGINS_PATH,
+                },
               },
             },
           },


docs/components.md
```diff
@@ -150,6 +150,32 @@ Kibana connects to the Elasticsearch cluster via the `elasticsearch-logging` Kub
...
+To install additional plug-ins, override the `plugins` item inside the `kibana` scope, like this:
+
+```jsonnet
+(import "../../manifests/platforms/aks.jsonnet") {
+    config:: import "kubeprod-autogen.json",
+    kibana+: {
+        plugins:: {
+            logtrail: {
+                version: "v0.1.31",
+                url: "https://github.com/sivasamyk/logtrail/releases/download/v0.1.31/logtrail-6.7.1-0.1.31.zip",
```
dbarranco: I think you can use in the url `$.version`.
If not that var won't be used

docs/components.md
```diff
@@ -150,6 +150,32 @@ Kibana connects to the Elasticsearch cluster via the `elasticsearch-logging` Kub
...
+            },
+            "analyze-api-ui-plugin": {
+                version: "6.7.1",
+                url: "https://github.com/johtani/analyze-api-ui-plugin/releases/download/6.7.1/analyze-api-ui-plugin-6.7.1.zip",
```
dbarranco: I think you can use in the url `$.version`.
If not that var won't be used

manifests/components/kibana.jsonnet
```diff
@@ -38,18 +39,69 @@ local strip_trailing_slash(s) = (
...
+    */
+  },
+
   es: error "elasticsearch is required",
 
   serviceAccount: kube.ServiceAccount($.p + "kibana") + $.metadata {
   },
 
+  pvc: kube.PersistentVolumeClaim($.p + "kibana-plugins") + $.metadata {
+    storage: "15Gi",
```
dbarranco: Isn't is this too much space for the plugins?

Also, if by default we are not going to install any plugin we could have less space assigned to that volume

manifests/components/kibana.jsonnet
```diff
@@ -38,18 +39,69 @@ local strip_trailing_slash(s) = (
...
       template+: {
         spec+: {
           securityContext: {
             fsGroup: 1001,
           },
+          volumes_+: {
+            plugins: kube.PersistentVolumeClaimVolume($.pvc),
+          },
+          initContainers_+: {
+            elasticsearch_plugins_install: kube.Container("elasticsearch-plugins-install") {
```
dbarranco: shouldn't this name be `kibana_plugins_install`?
 (just a small suggesstion: I would rename it like `kibana-plugin-installer`)

manifests/components/kibana.jsonnet
```diff
@@ -38,18 +39,69 @@ local strip_trailing_slash(s) = (
     },
   },
 
+  // List of Kibana plug-ins to install
+  plugins:: {
+    /* Example:
+    logtrail: {
+      version: "v0.1.31",
+      url: "https://github.com/sivasamyk/logtrail/releases/download/v0.1.31/logtrail-6.7.1-0.1.31.zip",
```
dbarranco: I think you can use in the url $.version.
If not that var won't be used

manifests/components/kibana.jsonnet
```diff
@@ -38,18 +39,69 @@ local strip_trailing_slash(s) = (
...
+                KIBANA_PLUGINS: std.join('\n', ["%(name)s,%(version)s,%(url)s" % plugin for plugin in plugins]),
+              },
+              command: [
+                "/bin/bash",
+                "-c",
+                |||
+                  set -e
+                  set -x
+                  exec &> >(tee /opt/bitnami/kibana/plugins/log)
+                  rm -rf /opt/bitnami/kibana/plugins/lost+found
```
dbarranco: Since this directory will be automatically re-created I'm not sure that it is worth to be constantly deleting it.

dbarranco:
left some questions, but overall lgtm.

nice stuff!

docs/components.md
falfaro: `version` is used in line 79 in the following expression:

```
                local plugins = [{name: name} + $.plugins[name] for name in std.objectFields($.plugins)],
                KIBANA_PLUGINS: std.join('\n', ["%(name)s,%(version)s,%(url)s" % plugin for plugin in plugins]),
```

docs/components.md
falfaro: `version` is used in line 79 in the following expression:

```
                local plugins = [{name: name} + $.plugins[name] for name in std.objectFields($.plugins)],
                KIBANA_PLUGINS: std.join('\n', ["%(name)s,%(version)s,%(url)s" % plugin for plugin in plugins]),
```

manifests/components/kibana.jsonnet
falfaro: Good point. I've reduced that to 2.

manifests/components/kibana.jsonnet
falfaro: Indeed!

manifests/components/kibana.jsonnet
falfaro: I dunno. I copied the code from the Helm chart. I believe it is deleted because being a directory, KIbana would get confused and believe it is a plug-in.

docs/components.md
```diff
@@ -150,6 +150,32 @@ Kibana connects to the Elasticsearch cluster via the `elasticsearch-logging` Kub
...
+
+Add-on functionality for Kibana is implemented with plug-in modules. Some known plug-ins for Kibana are listed here: https://www.elastic.co/guide/en/kibana/6.7/known-plugins.html.
+
+To install additional plug-ins, override the `plugins` item inside the `kibana` scope, like this:
+
+```jsonnet
+(import "../../manifests/platforms/aks.jsonnet") {
+    config:: import "kubeprod-autogen.json",
+    kibana+: {
+        plugins:: {
```
anguslees: Just in case we need to install our own plugins at some point:

```suggestion
        plugins+: {
```

manifests/components/kibana.jsonnet
```diff
@@ -38,18 +39,69 @@ local strip_trailing_slash(s) = (
...
+                local plugins = [{name: name} + $.plugins[name] for name in std.objectFields($.plugins)],
+                KIBANA_PLUGINS: std.join('\n', ["%(name)s,%(version)s,%(url)s" % plugin for plugin in plugins]),
+              },
+              command: [
+                "/bin/bash",
+                "-c",
+                |||
+                  set -e
+                  set -x
+                  exec &> >(tee /opt/bitnami/kibana/plugins/log)
```
anguslees: Is it expected that the logs are persisted here, rather than (only) in the k8s initcontainer log?
Unless there is a need/expectation that they will be here, I suggest we just log to stdout/stderr, as usual.

manifests/components/kibana.jsonnet
```diff
@@ -38,18 +39,69 @@ local strip_trailing_slash(s) = (
...
+                  rm -rf /opt/bitnami/kibana/plugins/lost+found
+                  for i in ${KIBANA_PLUGINS}
+                  do
+                    IFS=',' read -ra PLUGIN <<< "$i"
```
anguslees: This is fine, but if you want to avoid bash features for some reason (/bin/sh might be a lighterweight shell), then you could also do:
```
IPFS=,; set -- $PLUGIN; IPFS=
/opt/bitnami/kibana/bin/kibana-plugin remove --silent $1 || :
/opt/bitnami/kibana/bin/kibana-plugin install --no-optimize $3
```

manifests/components/kibana.jsonnet
```diff
@@ -38,18 +39,69 @@ local strip_trailing_slash(s) = (
...
+                    /opt/bitnami/kibana/bin/kibana-plugin remove --silent "${PLUGIN[0]}" || :
```
anguslees: If I'm reading this correctly, I think it is _always_ removing and reinstalling each plugin on every pod restart.  If we're going to do that, then we may as well not use a PVC?  Is there some way we can skip this if the plugin/version is already installed?

manifests/components/kibana.jsonnet
```diff
@@ -38,18 +39,69 @@ local strip_trailing_slash(s) = (
...
+
   deploy: kube.Deployment($.p + "kibana") + $.metadata {
     spec+: {
       template+: {
         spec+: {
           securityContext: {
             fsGroup: 1001,
           },
+          volumes_+: {
+            plugins: kube.PersistentVolumeClaimVolume($.pvc),
```
anguslees: This is a Deployment using a (ReadWriteOnce) PVC.  This will be ok, provided we only have a single replica (and kube.libsonnet already adjusts the update strategy accordingly for this situation).
Do we ever want multiple kibana replicas?

manifests/components/kibana.jsonnet
anguslees: I note the helm chart does (basically) the following to check if a plugin is already installed:
```sh
kibana-plugin list | grep -q $name@$version
```

Unlike the helm script, we want to just always make the installed plugin state equal to the user settings - the helm chart has a mix of "install/upgrade" and "reset".  I think I would do something like (untested):

```jsonnet
local wanted = std.join("\n",
  ["%s@%s,%s" % [k, $.plugins[k].version, $.plugins[k].url] for k in std.objectFields($.plugins)]);
|||
  set -e
  cd /tmp
  echo %s | sort > wanted.list 
  kibana-list | grep @  | sort > installed.list
  join -v2 -t, -j1 wanted.list installed.list |
  while read plugin; do
    kibana-plugin remove ${plugin%@*}
  done
  join -v1 -t, -j1 -o1.2 wanted.list installed.list |
  while read url; do
    kibana-plugin install --no-optimize $url
  done
||| % std.escapeStringBash(wanted)
```

(`join` is one of the most underappreciated shell commands :)

manifests/components/kibana.jsonnet
falfaro: To be honest, I don't see a good reason to have multiple Kibana replicas. But since the PVC is only used to keep the plug-ins, we could use a 1-1 mapping between a Kibana Pod and the PVC used for plug-ins, right?

manifests/components/kibana.jsonnet
falfaro: Nice. I've implemented your very nice suggestion and also tested it.

manifests/components/kibana.jsonnet
falfaro: Then once the init container is done, how do you access its logs?

docs/components.md
falfaro: Done

manifests/components/kibana.jsonnet
anguslees: > how do you access its logs?

`kubectl logs kibana-xxx -c kibana-plugins-install`

manifests/components/kibana.jsonnet
anguslees: sg, happy to leave it as-is based on the idea that we think this will only ever be replicas=1.  If we ever decide to change that, then we can revisit this at that time.

manifests/components/kibana.jsonnet
```diff
@@ -38,18 +39,68 @@ local strip_trailing_slash(s) = (
...
+              },
+              local wanted = std.join("\n", ["%s@%s,%s" % [k, $.plugins[k].version, $.plugins[k].url] for k in std.objectFields($.plugins)]),
+              command: [
+                "/bin/bash",
+                "-c",
+                |||
+                  set -e
+                  exec &> >(tee /opt/bitnami/kibana/plugins/log)
+                  rm -rf /opt/bitnami/kibana/plugins/lost+found
+                  echo %s | sort > /opt/bitnami/kibana/plugins/wanted.list
```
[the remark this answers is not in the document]
falfaro: Fixed.

manifests/components/kibana.jsonnet
```diff
@@ -38,18 +39,68 @@ local strip_trailing_slash(s) = (
...
+          },
+          initContainers_+: {
+            kibana_plugins_install: kube.Container("kibana-plugins-install") {
+              image: KIBANA_IMAGE,
+              securityContext: {
+                allowPrivilegeEscalation: false,
+              },
+              local wanted = std.join("\n", ["%s@%s,%s" % [k, $.plugins[k].version, $.plugins[k].url] for k in std.objectFields($.plugins)]),
+              command: [
+                "/bin/bash",
```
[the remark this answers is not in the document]
falfaro: Fixed.

falfaro:
@dbarranco @anguslees PTAL

falfaro:
Re-based and waiting for Jenkins... If it fails for whatever reason, I'm tempted to merge it manually.

sameersbn:
> Re-based and waiting for Jenkins... If it fails for whatever reason, I'm tempted to merge it manually.

Please don't. `kibana` enters a crashloop and therefore the tests are not passing. Please take a look at https://jenkins-bkpr.nami.run/blue/organizations/jenkins/BKPR%2Fkube-prod-runtime/detail/pr-530/2/pipeline 

`pod/kibana-6d4d49fc9b-dhjp6                     0/1     Init:CrashLoopBackOff   10         30m`

ref: https://jenkins-bkpr.nami.run/blue/organizations/jenkins/BKPR%2Fkube-prod-runtime/detail/pr-530/2/pipeline

manifests/components/kibana.jsonnet
```diff
@@ -38,18 +39,66 @@ local strip_trailing_slash(s) = (
...
+              command: [
+                "/bin/sh",
+                "-c",
+                |||
+                  set -e
+                  rm -rf /opt/bitnami/kibana/plugins/lost+found
+                  echo %s | sort > /tmp/wanted.list
+                  /opt/bitnami/kibana/bin/kibana-plugin list | grep @ | sort > /tmp/installed.list
+                  join -v2 -t, -j1 /tmp/wanted.list /tmp/installed.list | while read plugin; do
+                    ${plugin:+/opt/bitnami/kibana/bin/kibana-plugin remove "${plugin%%@*}"}
```
sameersbn: updated to account for empty list /cc @anguslees

manifests/components/kibana.jsonnet
```diff
@@ -38,18 +39,66 @@ local strip_trailing_slash(s) = (
...
+                  done
+                  join -v1 -t, -j1 -o1.2 /tmp/wanted.list /tmp/installed.list | while read url; do
+                    ${url:+/opt/bitnami/kibana/bin/kibana-plugin install --no-optimize "$url"}
```
sameersbn: updated to account for empty list /cc @anguslees

manifests/components/kibana.jsonnet
anguslees: ... aha!  I was confused because it _should_ have done-the-right-thing with empty lists.  The issue is/was that the "echo %s | sort >/tmp/wanted.list" would create a list with an empty line, because `echo ""`.

So another solution (with hindsight) would have been to do `echo %s | grep . | sort >/tmp/wanted.list` or a more sophisticated `echo` variation.  (Not a suggestion to change, I just wanted to understand the underlying issue)

PR #546, opened 2019-05-17
adding elasticsearch-curator

jbianquetti-nami:
This  PR provides a cronjob (based on `bobrik/docker-curator` container),  which executed on a daily basis, takes care of removing all indices with more than `60` days old.

### DIFF
--- a/manifests/components/elasticsearch-curator.jsonnet
+++ b/manifests/components/elasticsearch-curator.jsonnet
@@ -17,17 +17,17 @@
  * limitations under the License.
  */
 
-local kube = import "../lib/kube.libsonnet";
-local utils = import "../lib/utils.libsonnet";
+local kube = import '../lib/kube.libsonnet';
+local utils = import '../lib/utils.libsonnet';
 
-local CURATOR_IMAGE = (import "images.json").curator;
+local CURATOR_IMAGE = (import 'images.json').curator;
 
 // Implement elasticsearch-curator as a Kubernetes CronJob
 local elasticsearch_curator = {
-  namespace:: null,
-  name:: "elasticsearch-curator",
+  namespace:: 'kubeprod',
+  name:: 'elasticsearch-curator',
   retention_days:: 60,
-  elasticsearch_host:: "elasticsearch-logging",
+  elasticsearch_host:: 'elasticsearch-logging',
   elasticsearch_port:: 9200,
   elasticsearch_curator_config: kube.ConfigMap($.name) {
     metadata+: {
@@ -87,8 +87,8 @@ local elasticsearch_curator = {
           logformat: default
           blacklist: ['elasticsearch', 'urllib3']
       |||,
-      "action_file.yml": std.format(self.action_file_yml_tmpl, [$.retention_days]),
-      "config.yml": std.format(self.config_yml_tmpl, [$.elasticsearch_host, $.elasticsearch_port]),
+      'action_file.yml': std.format(self.action_file_yml_tmpl, [$.retention_days]),
+      'config.yml': std.format(self.config_yml_tmpl, [$.elasticsearch_host, $.elasticsearch_port]),
     },
   },
   elasticsearch_curator_cronjob: kube.CronJob($.name) {
@@ -96,19 +96,19 @@ local elasticsearch_curator = {
       namespace: $.namespace,
     },
     spec+: {
-      schedule: "30 0 * * *",
+      schedule: '30 0 * * *',
       jobTemplate+: {
         spec+:
           {
             template+: {
               spec+: {
                 containers_+: {
-                  curator: kube.Container("curator") {
+                  curator: kube.Container('curator') {
                     image: CURATOR_IMAGE,
-                    args: ["--config", "/etc/config/config.yml", "/etc/config/action_file.yml"],
+                    args: ['--config', '/etc/config/config.yml', '/etc/config/action_file.yml'],
                     volumeMounts_+: {
                       config_vol: {
-                        mountPath: "/etc/config",
+                        mountPath: '/etc/config',
                         readOnly: true,
                       },
                     },
--- a/manifests/components/elasticsearch-curator.jsonnet
+++ b/manifests/components/elasticsearch-curator.jsonnet
@@ -1,0 +1,128 @@
+/*
+ * Bitnami Kubernetes Production Runtime - A collection of services that makes it
+ * easy to run production workloads in Kubernetes.
+ *
+ * Copyright 2018-2019 Bitnami
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


jbianquetti-nami:
This PR would work for  issues/539

manifests/lib/kube.libsonnet
```diff
@@ -429,6 +429,38 @@
...
+        },
+      },
+      schedule: error "Need to provide spec.schedule",
+      successfulJobsHistoryLimit: 10,
+      failedJobsHistoryLimit: 20,
+      concurrencyPolicy: "Forbid",
+    },
+  },
+
+  JobSpec: {
```
jjo: I think we should reuse this `JobSpec` also for the `Job` object ?

manifests/components/elasticsearch-curator.jsonnet
```diff
@@ -0,0 +1,128 @@
...
+local utils = import '../lib/utils.libsonnet';
+
+local CURATOR_IMAGE = (import 'images.json').curator;
+
+// Implement elasticsearch-curator as a Kubernetes CronJob
+local elasticsearch_curator = {
+  namespace:: 'kubeprod',
+  name:: 'elasticsearch-curator',
+  retention_days:: 60,
+  elasticsearch_host:: 'elasticsearch-logging',
```
jjo: Wonder if we should leave these two instead as
```
error "elasticsearch_host must be externally provided ..."
```
then set it from the main ES object [*1]

manifests/components/elasticsearch.jsonnet
```diff
@@ -227,4 +230,5 @@ local ELASTICSEARCH_TRANSPORT_PORT = 9300;
       sessionAffinity: "None",
     },
   },
+  curator: elasticsearch_curator,
```
jjo: [*1] ie here do:
```
   curator: elasticsearch_curator {
        elasticsearch_host: $.svc.metadata.name,
        elasticsearch_port: ELASTICSEARCH_HTTP_PORT,
   }
```

jbianquetti-nami:
thanks!  /me is addressing comments

manifests/components/images.json
```diff
@@ -9,5 +9,6 @@
     "kibana": "bitnami/kibana:6.7.2-r1",
     "nginx-ingress-controller": "bitnami/nginx-ingress-controller:0.24.1-r4",
     "oauth2_proxy": "bitnami/oauth2-proxy:3.1.0-r1",
+    "curator": "bobrik/curator:5.7.6",
```
sameersbn: Would be great if you sort this entry in the list, it makes it easy to parse the list.

anguslees:
> Would it be possible to add a test for this change?

Hrm.  I think doing an end-to-end test requires logging something, setting up a really short cron timeout, and then observing that the earlier log is actually rotated away.  That sounds like it's going to necessarily be a fairly slow test :(  Any better ideas?

manifests/lib/kube.libsonnet
jbianquetti-nami: Addressed

manifests/components/elasticsearch-curator.jsonnet
jbianquetti-nami: Addressed

manifests/components/elasticsearch.jsonnet
jbianquetti-nami: Addressed

manifests/components/images.json
jbianquetti-nami: Addressed

jbianquetti-nami:
> > Would it be possible to add a test for this change?
> 
> Hrm. I think doing an end-to-end test requires logging something, setting up a really short cron timeout, and then observing that the earlier log is actually rotated away. That sounds like it's going to necessarily be a fairly slow test :( Any better ideas?

The tests I've in mind is:
- create X number of empty ES indices, using the pattern `logstash-YYYY.MMM.DD`, which is the default pattern. 
- Instantiate the curator to delete those indices with more than 90 days
- Count remaining indices on ES after a run of curator
- Drop all indices. 

@anguslees  @sameersbn   Can this do the trick?

anguslees:
> The tests I've in mind is:
> create X number of empty ES indices, using the pattern logstash-YYYY.MMM.DD, which is the default pattern.

Artificially creating old indices is an excellent idea - and better than my earlier suggestion of creating them and then waiting for them to "naturally" age out.

Assuming you go with your suggestion and artificially create old indices, please also add a trivial test that new/current indices are generated using the name-pattern you expect.  Otherwise that name could change and we wouldn't realise the "glue" between es and curator had been broken.

> I'm using `jsonnet fmt -i --string-style d ${file}`

I think our usual style is `jsonnet fmt --indent 2 --string-style d --comment-style s --no-pad-arrays --no-pad-objects --pretty-field-names`

I/we don't routinely use `jsonnet fmt` on these files - but we probably should.  You will see some deviations from the above fmt as a result.

If you can, I'd prefer it if you did _not_ re-fmt existing files as part of this PR - and then we can do a noop-fmt-only cleanup PR separately (and add a unittest so it stays formatted, etc).

manifests/components/elasticsearch-curator.jsonnet
```diff
@@ -0,0 +1,128 @@
...
+    },
+  },
+  elasticsearch_curator_cronjob: kube.CronJob($.name) {
+    metadata+: {
+      namespace: $.namespace,
+    },
+    spec+: {
+      schedule: '30 0 * * *',
+      jobTemplate+: {
+        spec+:
```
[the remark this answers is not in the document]
jbianquetti-nami: +1, addressed.

jbianquetti-nami:
> I think our usual style is `jsonnet fmt --indent 2 --string-style d --comment-style s --no-pad-arrays --no-pad-objects --pretty-field-names`

Then, we should  document it elsewhere and adding a test to reject PR that doesn't comply with the policy (for just the PR affected files)

manifests/components/elasticsearch-curator.jsonnet
```diff
@@ -0,0 +1,128 @@
...
+                },
+              },
+            },
+          },
+      },
+    },
+  },
+};
+
+kube.List() { items_+: elasticsearch_curator }
```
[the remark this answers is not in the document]
jbianquetti-nami: Addressed

manifests/components/elasticsearch-curator.jsonnet
```diff
@@ -0,0 +1,128 @@
...
+          timeout: 30
+          master_only: False
+
+        logging:
+          loglevel: INFO
+          logfile:
+          logformat: default
+          blacklist: ['elasticsearch', 'urllib3']
+      |||,
+      'action_file.yml': std.format(self.action_file_yml_tmpl, [$.retention_days]),
```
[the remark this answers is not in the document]
jbianquetti-nami: I'm having problems when dealing with `True/False` to `true/false` conversion.  
That piece of code is in Python, so we need to maintain as is. I've tried several approaches, but  I see there's little advantage on translating YAML  (with Pythonism) -> remove Pythonism -> jsonnet structure  ->  instantiate as YAML -> check that values are Python-friendly.  

There's not a big win, so I'm not totally sure of doing this... 

@jjo What's your take on this?

jbianquetti-nami:
@sameersbn Can you take a look at the test? I've updated and simplified the test as we agreed.

sameersbn:
bors retry

PR #554, opened 2019-05-23
Enforce a jsonnet fmt style

anguslees:
Add `Makefile` support for running and enforcing `jsonnet fmt` style
against our jsonnet files.

`make -C manifests fmttest` does a read-only verification.  This is
now called from `Jenkinsfile` to enforce style compliance.

`make -C manifests fmt` will update files in-place to follow the style
guide.

Fwiw, the current change (and jenkins config) will use `jsonnet`
0.12.1.  There have been minor `fmt` style deviations over the history
of `jsonnet`.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -327,8 +327,11 @@ spec:
                             stage('Manifests') {
                                 withGo() {
                                     dir("${env.WORKSPACE}/src/github.com/bitnami/kube-prod-runtime/manifests") {
-                                        withEnv(["PATH+KUBECFG=${tool 'kubecfg'}"]) {
-                                            sh 'make validate KUBECFG="kubecfg -v"'
+                                        withEnv([
+                                            "PATH+KUBECFG=${tool 'kubecfg'}",
+                                            "PATH+JSONNET=${tool 'jsonnet'}",
+                                        ]) {
+                                            sh 'make fmttest validate KUBECFG="kubecfg -v"'
                                         }
                                     }
                                 }
--- a/manifests/Makefile
+++ b/manifests/Makefile
@@ -1,4 +1,15 @@
 KUBECFG = kubecfg
+JSONNET_FMT = jsonnet fmt
+
+FMTFLAGS = \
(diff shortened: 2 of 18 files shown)


anguslees:
NB: this PR uses:
```make
FMTFLAGS = \
	--indent 2 \
	--string-style d \
	--comment-style s \
	--no-pad-arrays \
	--no-pad-objects \
	--pretty-field-names \
	--max-blank-lines 2 \
	--sort-imports
```

If we want something different, now is the time to discuss that :)  (Or we can change later of course.  It's trivial to re-run `make fmt` with different flags at any time)

I think the variants worth discussing are:
- preserve our existing style (this is mostly my personal style, derived from being json-like (double quote), and compact (minimal internal whitespace), because k8s configs are already crazy-indent-verbose).
- same, with pad-arrays and/or pad-objects
- upstream jsonnet fmt defaults (single-quote, pad-arrays, pad-objects)

jbianquetti-nami:
I've not a strong position about formatting on the aesthetical side, it works for me right now and I agreed with your proposed formatting. 

But, (yes, always there's one of this) the point of standardizing formatting to minimum default `fmt -i`  is to talk a common "slang" on different repositories or projects.   My 2 cnts.

mkmik:
I prefer the default formatting. It's ugly, but it gets less in the way to people who have autoformat on save feature in their editors/IDEs.

If I have time I'd like to contribute upstream the support for a `.jsonnetfmt` file (like `.clangformat`) at the root of a dir-tree, or a `// -*- jsonnet-fmt: ... -*-` comment.

jjo:
> I prefer the default formatting. It's ugly, but it gets less in the way to people who have autoformat on save feature in their editors/IDEs.

+1 on using jsonnet fmt default, specifically on the side of removing bars(/specialized setup)
from contributors

dbarranco:
I also don't have a strong opinion on this, but I think I'm more on the side of the default formatting of jsonnet.
Thanks for the PR

PR #727, opened 2020-02-20
ci: Use the Authorization HTTP header for GitHub API queries, as using the `access_token` query parameter is deprecated

sameersbn:
ref: https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param/


sameersbn:
bors merge

PR #744, opened 2020-03-05
docs: document oauth2 annotations in application developers guide

### LINKED ISSUE #734: Documentation for OAuth2 annotations
You can use the same single-sign on in your apps as the metrics/log stack provided by BKPR. You just have to add 2 annotations:

```
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
  name: ingress
  annotations:
    nginx.ingress.kubernetes.io/auth-signin: https://auth.$(DOMAIN)/oauth2/start?rd=%2F$server_name$escaped_request_uri
    nginx.ingress.kubernetes.io/auth-url: https://auth.$(DOMAIN)/oauth2/auth
```

But this feature doesn't seem to be documented anywhere. It would be good to be able to easily find it in /docs.

### LINKED ISSUE #735: Ingress documentation refers to wrong group name
The [application developer docs](https://github.com/bitnami/kube-prod-runtime/blob/master/docs/application-developers-reference-guide.md) have an example ingress that uses `extensions/v1beta1` but the [official Kubernetes documentation](https://kubernetes.io/docs/concepts/services-networking/ingress/) uses `networking.k8s.io/v1beta1`. I guess the latter is correct. Both work when all you are doing is `kubectl apply` but I'm guessing there might be tools out there that assume it is one way and not the other.

sameersbn:
- Closes #734
- Closes #735

### DIFF
--- a/docs/application-developers-reference-guide.md
+++ b/docs/application-developers-reference-guide.md
@@ -43,6 +43,24 @@ spec:
             servicePort: 80
 ```
 
+## Restricting access with OAuth Authentication
+
+BKPR installs a [OAuth2 Proxy](https://github.com/pusher/oauth2_proxy/) for restricting access to the BKPR dashboards.
+
+Externally accessible web applications deployed by users on the cluster are not protected by this OAuth scheme. However you can easily enable this by adding the following annotations to your applications `Ingress` resources.
+
+```yaml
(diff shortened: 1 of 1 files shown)


sameersbn:
cc @dsyer

PR #814, opened 2020-05-20
[jjo] feat: all clouds tested for 1.15-1.16, fix integration tests

jjo:
* all 3 platforms now tested for 1.15 and 1.16 (GA on all),
  using 1.16 also for `generic`
* use same lock for `generic` and `gke` as they are both using
  GKE, this seems to have been the cause of flaky e2e testing results
* pin ginkgo CLI build to match `go mod` in `tests/`
* workaround https://issues.jenkins-ci.org/browse/JENKINS-41929
  to avoid the need for "2nd job run" when releases to test were changed
  from parameters properties (else it's results are confusing as they
  would stick to previous settings)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -192,7 +192,8 @@ def runIntegrationTest(String description, String kubeprodArgs, String ginkgoArg
 
                 withGo() {
                     dir("${env.WORKSPACE}/src/github.com/bitnami/kube-prod-runtime/tests") {
-                        sh 'go get github.com/onsi/ginkgo/ginkgo'
+                        // NOTE: ginkgo version pinned to the one used in tests/
+                        sh 'env GO111MODULE=on go get github.com/onsi/ginkgo/ginkgo@v1.12.0'
                         try {
                             sh """
                             ginkgo -v \
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,9 +18,11 @@ properties([
   // See releasesFromStr() function below on how we parse the _REL string
   parameters([
     stringParam(name: 'AKS_REL', defaultValue: '1.15,1.16', description: 'AKS releases to test (comma separated)'),
-    stringParam(name: 'EKS_REL', defaultValue: '1.14,1.15', description: 'EKS releases to test (comma separated)'),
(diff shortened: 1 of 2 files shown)


jjo:
bors retry

jjo:
bors try

jjo:
bors try

jjo:
@dbarranco @jbianquetti-nami @javsalgar PTAL, as I changed several other bits,
see my updated 1st comment, as you can see, several build runs in green already.

Jenkinsfile
```diff
@@ -10,20 +10,35 @@
...
-    stringParam(name: 'AKS_REL', defaultValue: '1.15,1.16', description: 'AKS releases to test (comma separated)'),
-    stringParam(name: 'EKS_REL', defaultValue: '1.14,1.15', description: 'EKS releases to test (comma separated)'),
-    stringParam(name: 'GKE_REL', defaultValue: '1.15,1.16-pre', description: 'GKE releases to test (comma separated)'),
-    stringParam(name: 'GEN_REL', defaultValue: '1.15', description: 'Generic-cloud releases to test (comma separated)'),
+    stringParam(name: 'AKS_REL', defaultValue: '', description: "Override AKS releases to test"),
+    stringParam(name: 'EKS_REL', defaultValue: '', description: "Override EKS releases to test"),
+    stringParam(name: 'GKE_REL', defaultValue: '', description: "Override GKE releases to test"),
+    stringParam(name: 'GEN_REL', defaultValue: '', description: "Override Generic-cloud releases to test"),
+    stringParam(name: 'TIMEOUT', defaultValue: '150', description: 'Full e2e tests run timeout in minutes'),
+    stringParam(name: 'PAUSE_TIME', defaultValue: '15', description: 'Time to pause on failure in minutes (before deletion)'),
```
dbarranco, line 34: re-lgtm, thanks for the diff.

Ooc, which is the use case of the releases overrides?

jjo: > Ooc, which is the use case of the releases overrides?

E.g. manually testing a new release, or looping on an specific one (ie removing others) while debugging (this latter was my use case, focusing on `generic` build issues we had).

PR #858, opened 2020-06-25
[jjo] fix: support Azure increasingly sized auth cookies, verified to work with latest oauth2-proxy:5.1.x image

jjo:


jjo:
approved and tested, let's

bors r+

PR #884, opened 2020-07-24
[jjo] fix grafana dashboard imports

jjo:
Fix grafana.jsonnet dashboard imports to use full vendored
path instead of symlink, as the latter is not present at
`releases.kubeprod.io` cloudfront bucket (we ignore symlinks
when pushing manifest tree from src to the s3 bucket).


jjo:
bors merge

PR #935, opened 2020-09-22
Fix: Update Custodian version to custom Bitnami container and fix/improve AWS / Google policies

dbarranco:
We have been facing some more issues related with the clean-ups in AWS related to the AWS dependency errors. 
This means that if you want to delete a VPC you must delete before:

* EC2 instances
* Security groups
* NatGateways
* Elastic IP addresses
* InternetGateways

Otherwise the AWS API will throw a dependency error.

In this PR you will also see that I am setting the 0.9.6.2 version of the Cloud-Custodian container. This version contains the latest improvements I landed in the Cloud-Custodian repository **and the `detach` operation coded for the InternetGateway**, as it is required prior deleting the VPC.
I will contribute to the project with that operation, but I wouldn't like to block all the clean-ups in the meantime.

****

Apart from the fix of the AWS policies (for hostedzones / IAM users) I have enabled the Google Cloud policies to clean-up the GKE clusters.

### DIFF
--- a/jenkins/cloud-custodian/policies/aws.yaml
+++ b/jenkins/cloud-custodian/policies/aws.yaml
@@ -33,7 +33,7 @@ policies:
 - name: bkpr-delete-nat-gateways
   resource: aws.nat-gateway
   comment: |
-    Clean-up created NatGateways created by CFN
+    Clean-up NatGateways created by CFN
     for the BKPR continuous integration tests
   filters:
     - type: value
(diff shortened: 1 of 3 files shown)


jenkins/cloud-custodian/policies/aws.yaml
```diff
@@ -43,15 +95,28 @@ policies:
...
+  actions:
+    - type: delete
+
+- name: bkpr-delete-security-groups
   resource: aws.iam-user
   comment: |
-    Clean-up temporary serviceaccounts created by Jenkins-BKPR
-    for its continuous integration tests
+    Clean-up temporary internet gateways created by Jenkins-BKPR
+    for it's continuous integration tests
```
jbianquetti-nami: little typo

jjo (approved):

jbianquetti-nami (approved):

javsalgar (approved):

jbianquetti-nami (approved):
thanks for this

Issue #1111, opened 2021-02-23
Cannot install with git-for-windows: config path broken


WolfByttner:
Kubeprod will not correctly resolve the path to `kubeprod-autogen.json`. When I try, I get various strange error messages or weird file paths. I have attempted all file format paths that I can think of (see below for outputs).

**Steps to reproduce**

1. Setup a kubernetes cluster (I use GKE).
2. Build kubecfg for windows.
3. Add all required binaries to your path so Windows can see them.
4. Run the test script in git-for-windows (commonly known as "git bash") with your credentials and a path to kubeprod-autogen.json

Test script:
```
gcloud auth activate-service-account my-serviceaccount@my-project.iam.gserviceaccount.com --key-file=./gke-credentials.json

gcloud config set account my-serviceaccount@my-project.iam.gserviceaccount.com

export GCLOUD_ZONE="my-zone"

export BKPR_DNS_ZONE="my-cluster.my-domain.com"
export GCLOUD_USER="$(gcloud info --format='value(config.account)')"
export GCLOUD_PROJECT="my-project"
export GCLOUD_AUTHZ_DOMAIN="my-domain.com"
export GCLOUD_K8S_CLUSTER="my-cluster-name"

kubectl create clusterrolebinding cluster-admin-binding \
  --clusterrole=cluster-admin \
  --user=${GCLOUD_USER}


export GCLOUD_OAUTH_CLIENT_KEY="XXXXXXXXXXXXXXXXXXX"
export GCLOUD_OAUTH_CLIENT_SECRET="XXXXXXXXXXXXXXXXXXXXXXXXX"

kubeprod install gke \
  --email "${GCLOUD_USER}" \
  --dns-zone "${BKPR_DNS_ZONE}" \
  --project "${GCLOUD_PROJECT}" \
  --oauth-client-id "${GCLOUD_OAUTH_CLIENT_KEY}" \
  --oauth-client-secret "${GCLOUD_OAUTH_CLIENT_SECRET}" \
  --authz-domain "${GCLOUD_AUTHZ_DOMAIN}" \
  --kubeconfig "${KUBE_CONFIG}" \
  --config ${MY_FILEPATH}
```

Let us start with no --config flag. You get the following output.

```
$ ./install-kubeprod.sh
Installing platform gke
Error: open /C:\Users\wolfb\test/kubeprod-autogen.json: The filename, directory name, or volume label syntax is incorrect.
```

[...]

PR #1112, opened 2021-02-24
Fix: Make kubeprod resolve paths correctly on Windows

WolfByttner:
This is a fix that makes kubeprod correctly resolve `kubeprod-autogen.json` and the manifest files, as mentioned in https://github.com/bitnami/kube-prod-runtime/issues/1111. The fix works on Windows (in Git for Windows) and I can execute a full deployment.

I am partly raising this to see if the community has better suggestions for how to fix the issue. This is also for bookkeeping - we now have one documented fix to work from.

### DIFF
--- a/kubeprod/cmd/install.go
+++ b/kubeprod/cmd/install.go
@@ -22,6 +22,7 @@ package cmd
 import (
 	"fmt"
 	"regexp"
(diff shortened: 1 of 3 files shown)


javsalgar:
bors retry

PR #1317, opened 2021-06-15
Testing hypothesis on parallel AKS testing

scottmuc:
My hypothesis is that there may exist issues testing AKS in parallel.
There may exist a shared resource that results in the first env to
"obtain" this resource will succeed and the subsequent environment will
fail. Cheap to test this, so why not...

I decreased the number of K8s versions to test to minimise the length of
the pipeline and the possibility for random failures.

GEN_REL bumped to 1.19, though I'm not sure if that makes a difference.
Going to play with that setting to understand it more.


scottmuc:
Closing since this shows that a non-parallel run of AKS testing still results in the same failures in AKS 1.19

